PR #1, opened 2018-09-20
Concrete semantics

benozol:
There are a number of subtleties in the semantics (surprise!). I have marked the ones I considered important with `TODO`, and added some ideas in comments. Your option would be most welcome. What about using the Github comments for replies?


src/concrete/semantics.mlw
```diff
@@ -0,0 +1,263 @@
+module Mode
```
benozol: We should ask Github to support Why3 syntax!

src/concrete/semantics.mlw
```diff
@@ -0,0 +1,268 @@
...
+  *)
+  with eval_exprs bool input state expressions (list string) irreversable bool =
+
+    | eval_exprs_nil: forall prev inp sta.
+      eval_exprs prev inp sta Nil Nil sta.irr prev
+
+    | eval_exprs_cons: forall prev inp sta e sp es s irr1 irr2 bo b l1 l2 l3.
+      eval_expr inp sta e s irr1 bo ->
+      l1 = (if sp then split s else Cons s Nil) ->
+      eval_exprs (get_option bo previous) inp {sta with irr=irr1} es l2 irr2 b ->
```
benozol: Should `inp.previous`  be `get_option bo previous` as well for `eval_exprs`?

```shell
$ bash --posix -c 'false; x=A/$?/$(exit 2)$?$(exit 3); echo $x/$?'
A/1/2/3
```

src/concrete/semantics.mlw
```diff
@@ -0,0 +1,352 @@
...
+  | eval_esubshell: forall inp ctx ctx' sta sta' stm out s b.
+    eval_stmt inp ctx sta stm sta' ctx' {out with stdout = Stdout s} ->
+    b = Some (bool_of_behaviour out.behaviour) ->
+    eval_expr inp ctx sta (ESubshell stm) sta' s b
+
+    (* TODO EConcat, but decide first syntax for concat *)
+
+  (** Evaluation of expressions to a list of strings:
+
+      \[ (Bool, I, C, S) es \Downarrow (S, StringList, Bool) \]
```
benozol: Do we really need a behaviour for expression**s**?

src/language/syntax.mlw
```diff
@@ -1,14 +1,36 @@
...
+
+  use export Variable
+
+  type split = Split | NoSplit
   
   type expression =
     | ELiteral string
     | EVariable variable
     | ESubshell statement
     | EConcat expression expression
```
Niols: Two possibilities: either we put concatenation in `expression` or we consider expression lists. The first one gives a cleaner AST imho. The second one avoid having to say that the semantics for `EConcat` is associative and commutative.

src/language/syntax.mlw
```diff
@@ -1,14 +1,36 @@
...
 
+  (* TODO `$<n>`, `$@`, `$?` better explicit (`EArg int | EArgs | EPrevious`) than
+     as normal variable? Can be differentiated while parsing, right? `$?` is boolean,
+     however. How to separate? *)
```
Niols: Yes, `$?` will be a small problem. Because it is an integer in Shell and will be a boolean in CoLiS, which means that we have to be a bit careful when it is used.

`$@` will also be a problem as it has really special semantics.

src/language/syntax.mlw
```diff
@@ -1,14 +1,36 @@
...
+
+  (* TODO Or concat as flat list of expressions in a separate type instead? Like this
+
+     type scomp = Literal string | Variable var | Subshell stmt (* evaluates to string *)
+     type sexpr = Concat (list component) (* evaluates to string *)
+     type sexprs = list (sexpr, split) (* evaluates to string list *)
+
+     This would result in a more canonical representation.
+  *)
```
Niols: I tend to agree with that, the semantics would probably be clearer for readers. A few questions, though:
- do you really think we need a constructor in `sexpr`?
- would you rather use short names `scomp`, `sexpr`, etc. or long names `component`, `expression`, etc.? It's more a question about style, but it would be nice to decide that early and stay consistent. Although it makes things more verbose, I like using the full names as I find that easier to read after a while. But both are fine, really.
- how can we make clear the fact that `expressions <> expression list` (and I don't mean in the syntax but in the semantics where there is an important difference). Maybe we could find a different name for `expressions`?

src/language/syntax.mlw
```diff
@@ -17,16 +39,20 @@ module Syntax
...
     | SPipe statement statement
     | SWhile statement statement
     | SNoOutput statement
     | SForeach variable expressions statement
-    | SNot statement
     | SCall string expressions
     | SExit return_code
-end
+    (** CM: TODO `exit $ret` not representable? Add `CVar var` to `return_code`? `SExit
+        sexpr` instead with `$?` as expression? Stuck when sexpr not numerical/boolean? *)
```
Niols: `exit $ret` is rather rare, but it would be nice if we were able to represent it. I don't really know how, though. Either we add a boolean type to variables and we track when integers in Shell are used to represent booleans and translate that to CoLiS accordingly. Or we do as in Shell and encode booleans in integers. Then `SExit` could take any expression as long as it evaluates to an integer.

src/language/mystring.mlw
```diff
@@ -0,0 +1,12 @@
+module String
```
Niols: Do we really need more than just a type for strings?

src/concrete/semantics.mlw
```diff
@@ -0,0 +1,352 @@
+module Mode
+  type mode = Strict | Unstrict
```
Niols: `Unstrict`?

src/concrete/semantics.mlw
```diff
@@ -0,0 +1,352 @@
...
+  use option.Option
+
+  use Mode
+
+  (* No return-behaviour behaviour since we do not implement functions (for now) *)
+  type behaviour =
+    | BNormal bool
+    | BExit bool
+
+  let function behaviour_of_bool : mode -> bool -> behaviour =
```
Niols: Maybe we could find a better name than `*_of_bool`? It does actually much more than that as it takes a success/failure and decides which behaviour to give it depending on the mode.

src/concrete/semantics.mlw
```diff
@@ -0,0 +1,352 @@
...
+      end
+
+  let function behaviour_of_bool_opt : mode -> option bool -> behaviour =
+    fun strict b ->
+      match b with
+      | Some b' -> behaviour_of_bool strict b'
+      | None -> BNormal True
+      end
+
+  let function bool_of_behaviour : behaviour -> bool =
```
Niols: Same remark here, it's kind of "catching" the behaviour to tell us if it's a success or a failure.

src/concrete/semantics.mlw
```diff
@@ -0,0 +1,352 @@
...
+  use bool.Bool
+
+  use mystring.String
+  use syntax.Variable
+
+  use Mode
+  use Filesystem
+  use Behaviour
+
+  (* Better implement and use a buffer? *)
```
Niols: I think so, yes. We just have to be sure that the buffer has enough information for the proof obligations to be easy. In particular, it could be nice to have an easy way to write `exists s. stdout = (old stdout) ^ s and ...`. Maybe something like `let s = cutprefix (old stdout) stdout`?

src/concrete/semantics.mlw
```diff
@@ -0,0 +1,352 @@
...
+    under_condition = False;
+  }
+
+  (** {3 Read/write context for the evaluation of a term} *)
+
+  (** The state is unreversable, i.e. following a linear logic, each internal state should
+      be consumed exactly once *)
+  type state = {
+    filesystem: filesystem;
+    stdin: stdin; (* : @NJ How is this used in Debian scripts? Required interface for stdin? *)
```
Niols: It's not really used in Debian script. Dpkg does not send anything through stdin. The only ways it could happen are:
- when a script wants to interact with the user. However, we don't modelize that, and through stdin/stdout is not the preferred way
- in compound commands using pipes. Like `dpkg -L package | while read line; do ...`

src/concrete/semantics.mlw
```diff
@@ -0,0 +1,352 @@
...
+      state S', context C', and output O:
+
+      \[ (I, C, S) s \Downarrow (S', C', O) \]
+   *)
+
+  (** Universal variables for results of function applications and constructors in the
+      inductive definition below can be replaced by let-bindings when the following bug
+      report is resolved:
+
+      https://gitlab.inria.fr/why3/why3/issues/185
```
Niols: It is apparently closed on GitLab now

src/concrete/semantics.mlw
```diff
@@ -0,0 +1,352 @@
...
+   *)
+
+  inductive eval_stmt input context state statement state context output =
```
Niols: Is it on purpose that the inputs and outputs are in reverse order?

src/concrete/semantics.mlw
```diff
@@ -0,0 +1,352 @@
...
+    eval_expr inp ctx sta e sta' s b ->
+    ctx' = {ctx with env = ctx.env[v <- s]} ->
+    out = {stdout = empty_stdout; behaviour = behaviour_of_bool_opt (strict inp) b} ->
+    eval_stmt inp ctx sta (SAssignment v e) sta' ctx' out
+  
+  | eval_sequence_normal: forall inp sta sta1 sta2 s1 s2 ctx ctx1 ctx2 out1 out2 out3 b1.
+    eval_stmt inp ctx sta s1 sta1 ctx1 {out1 with behaviour = BNormal b1} ->
+    eval_stmt {inp with previous = b1} ctx1 sta1 s2 sta2 ctx2 out2 ->
+    out3 = seq_outputs out1 out2 ->
+    eval_stmt inp ctx sta (SSequence s1 s2) sta2 ctx out3
```
Niols: It is kinda hard to read with the short names. But I'm also afraid of using full names as it because veeeery verbose:

    | eval_sequence_normal: forall input state state1 state2 statement1 statement2 context context1 context2 output1 output2 output3 behaviour1.
      eval_statement input context state statement1 state1 context1 {output1 with behaviour = BNormal behaviour1} ->
      eval_statement {input with previous = behaviour1} context1 state1 statement2 state2 context2 output2 ->
      output3 = seq_outputs output1 output2 ->
      eval_statement input context state (SSequence statement1 statement2) state2 context output3

I'm not even sure it's easier to read.

I also wonder if we should maybe rename `statement` in something else in order to improve readability by removing the confusion between "state" and "statement". For now, the names are rather good because they don't clash, even on the first letter: input, output, context, state, behaviour, etc. Maybe we should go for something like "command", "term", or something else?

src/concrete/semantics.mlw
Niols: Considering where they are used, maybe not? We need them for expression (without s), for sure. For instance `a=$(false)` has a behaviour `Exit False`.

However, it is not the case in expressions: `echo $(false) a b` is just the same as `echo a b`. If we are convinced that it is not needed, it would be great to remove that. Every simplification of the semantics are welcome.

src/concrete/semantics.mlw
```diff
@@ -0,0 +1,352 @@
...
+
+      \[ (Bool, I, C, S) es \Downarrow (S, StringList, Bool) \]
+
+      CM: reuse `input.previous` instead of `option bool` or at least `bool` with default
+      `True`
+
+      BB: Indeed, even shell is using something like input.previous for this - the
+      following line prints "A/123-B/0" !
+
+      bash --posix -c 'x=$(exit 123)$(echo A/$?)-$(echo B/$?); echo $x'
```
Niols: Wait. Does that mean we need a behaviour for expressions because it has an impact on the following expressions?

Niols:
I added a lot of comments that just popped into my mind. Feel free not to take them into account when I just say nonsense :-)

src/concrete/semantics.mlw
benozol: Maybe `inject_behaviour : mode -> bool -> behaviour` which clarifies that it’s not just a simple conversion.

src/concrete/semantics.mlw
benozol: Idea: `stdin` and `stdout` maybe as `list string` to simplify the specification of `read`
Hypothesis: `echo -n` only for user-interaction, never under `$()` or '|'

src/concrete/semantics.mlw
```diff
@@ -0,0 +1,352 @@
...
+  let constant empty_state = {
+    filesystem = empty_filesystem;
+    stdin = empty_stdin;
+  }
+
+  (** The context of the interpretation, resulting contexts may be ignored (e.g., from
+      subprocesses) *)
+  type context = {
+    env: map variable string
+  }
```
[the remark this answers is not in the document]
benozol: Indeed! Any “language-context” goes here.

src/concrete/semantics.mlw
Niols: Probably true as far as I know

Niols: eg

```ocaml
let echo (n: bool) (line: string) (stdout: string list) : string list =
    match stdout with
    | [] -> assert false
    | h :: t ->
        let stdout = (h ^ line) :: t in
        if n then stdout else "" :: stdout
```

src/concrete/semantics.mlw
```diff
@@ -0,0 +1,352 @@
...
+      if notb (inp.under_condition)
+      then Strict
+      else Unstrict
+
+  (** Combine a sequence of two outputs, ignoring the behaviour of the first output *)
+  let function seq_outputs : output -> output -> output =
+    fun out1 out2 ->
+      match out1, out2 with
+      | {stdout = Stdout s1}, {stdout = Stdout s2; behaviour = b} ->
+        {stdout = Stdout (s1 ^ s2); behaviour = b}
```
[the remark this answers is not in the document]
Niols: We know that we always use it when out1.behaviour is normal.
Also, it is only for the semantics, so having a type for buffers does not change a thing.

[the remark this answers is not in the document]
benozol: TODO invariant that `out1.behaviour = BNormal _`

src/concrete/semantics.mlw
```diff
@@ -0,0 +1,352 @@
...
+  let function get_option opt default =
+    match opt with
+    | Some v -> v
+    | None -> default
+    end
+
+  (** Interprete a command defined in the document *Specification of UNIX Commands*.
+
+      It acts upon the input and irreversable state, and returns a stdout, a behaviour
+      (True/False), and another irreversable state *)
```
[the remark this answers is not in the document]
Niols: That's right.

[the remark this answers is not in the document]
Niols: Except for the order of the arguments.

src/concrete/semantics.mlw
Niols: dash does not have the same behaviour on that example. And the standard is not really clear in that regard. It only says "$? expands to the decimal exit status of the most recent pipeline (see Pipelines)."

src/concrete/semantics.mlw
benozol: TODO remove behaviour for expression**s**

src/language/syntax.mlw
Niols: We'll see

src/language/syntax.mlw
benozol: * `$@` is not required for now (no functions, no `shift`)
* `$@` could be added by replacing `Call name exprs` with `Call exprs` and defining `type exprs = list ((sexpr, split) | $@` (pseudocode)
* the callable will not be statically known any more

src/language/syntax.mlw
Niols: Indeed, very rare: 34 packages use variables in exit.

Niols: And 17 of return.

benozol: TODO
* remove `$?` (and `input.previous`) because it is used rarely in `exit`/`return`, and using it in a string and acting on it conflicts with the CoLiS hypothesis that behaviours are boolean.
* Then `SExit statement` with `SExit (SCall ["true"])` for `exit 0`

src/language/syntax.mlw
Niols: - `string_component`, `string_expression` and `list_expression`? That way, we have the semantics embedded in the name in the syntax.

src/concrete/semantics.mlw
Niols: Maybe `instruction`?

src/language/syntax.mlw
Niols: These comments should maybe go in the file itself? Or in a README file? We'll loose it if we close the PR

src/concrete/semantics.mlw
benozol: Good!

PR #2, opened 2018-09-28
Use Dune instead of OCamlBuild

Niols:
To be honest, I'm not so convinced anymore. I was thinking that we could use Dune to separate each directory in its specific module, making everything more modular, cleaner, more maintainable. But because of Why3's extraction mechanism, I'm not convinced that we can do it that easily. I see three solutions:

- We stick to ocamlbuild.
- We use Dune as if all the subdirectories of `src/` where one. This is the solution suggested in this PR.
- We keep the separation of the subdirectories, we use Dune. But in order to make it work, we have to put all the Why3 files in the same directory (`src/why3`?). I'm not sure how clean it can be, though.

Also, if the constraints turn out to be really independent from everything else, they should probably go into an other repository; we'll see that in time.

Any opinions?

### DIFF
--- a/src/dune
+++ b/src/dune
@@ -1,7 +1,7 @@
 (include_subdirs unqualified)
 
-(library
- (name symbolic)
- (public_name symbolic)
+(executable
+ (name concrete_colis)
+ (public_name colis)
  (libraries why3.extract)
  (flags :standard -w -27))
--- a/src/language/dune
+++ b/src/language/dune
@@ -1,2 +1,4 @@
+(include_subdirs unqualified)
+
 (ocamllex colisLexer)
 (menhir (modules colisParser))


Makefile
```diff
@@ -1,11 +1,23 @@
...
+	dune build @install
+
+test: build
+	dune test
+
+clean: clean-why3
+	dune clean
+
+extract-why3:
+	mkdir -p src/why3
```
benozol: Extra fun-fact is that previously extracted files from why3 modules that do not exist anymore are preserved, which may be confusing. What about `rm -f src/why3/*`?

src/dune
```diff
@@ -0,0 +1,8 @@
+(include_subdirs unqualified)
+
+(library
+ (name symbolic)
+ (public_name symbolic)
+ (libraries zarith)
```
benozol: Should be `why3.extract`? This has `zarith` as dependency.

src/concrete/concrete_colis.ml
```diff
@@ -1,34 +0,0 @@
-open Format
```
benozol: This was a check how the different parts play together (for now). Didn't compile?

benozol:
> We use Dune as if all the subdirectories of `src/` where one. This is the solution suggested in this PR.

I guess this is fine for now.

> But because of Why3's extraction mechanism…

Extracting why3 files to ml files with modules (as you mentioned via email sound) reasonable to me. This would also keep the module names sane (not `X__Y`). May be worth a feature request to why3 ?

benozol:
BTW I have merged #1 into master (including `NOTES.org`).

Makefile
Niols: Yes, this is rather surprising as a default behaviour. Your solution seems rather nice. Or else, after the extraction, `rm -f src/why3/*.bak`?

src/dune
Niols: Indeed, thanks

src/concrete/concrete_colis.ml
Niols: This is strange, I don't remember doing that

Niols: So what was this file for?

benozol: It was the skeleton of a main program for the concrete interpretation of a program, using the Moresmall or the Colis parser

Makefile
benozol: Oh maybe better just `rm -rf src/why3` before `mkdir src/why3`?

The `.bak` files aren’t a problem. But if your why3 file `file` contained a module `X` before the last extraction and you delete the module `X`, the ml file `test__X.ml` will still be here after a new extraction.

src/concrete/concrete_colis.ml
Niols: I really don't know how this file could disappear there… I'll put it back

Niols:
The only annoying thing that prevent from compiling here is:

```
File "src/why3/interpreter__Interpreter.ml", line 52, characters 14-28:
Error: Unbound value interp_builtin
```

Makefile
Niols: Yes, I realised that about the `.bak` files, that's why I chose your solution. I try not to use `rm -rf` but that would indeed achieve the same result. I don't know what is better.

Niols:
But that's not really related to Dune. I guess it's independent from this PR and could be fixed after the merge (if the PR looks good to you, don't hesitate to merge it)

benozol:
> build an executable instead of a library 

Good, that’s what I was just wondering about :)

I rebased branch `concrete-interpreter` on this branch. But the module `FromColis` is untracable for dune and I am puzzled. Would you mind having a look?

Niols:
Yes, I know why. I see the problem. I actually think that this is a problem in Dune. I'll push a work-around for now where the lexer and parser will be in `src/` and not in `src/language` while I open an issue in Dune's GitHub repository.

Niols:
I just did. However, it does not compile because of stuff in `concrete_colis.ml`.

benozol:
Thanks!

Niols:
OK, found it and pushed it:

- the `(ocamllex ...)` stanza wants to be in the same file as the module. It does not accept qualified module names (that is, with a path as a prefix)
- the `(menhir ...)` stanza wants to be in the same file as the `(executable ...)` or `(library ...)` stanza. It does accept qualified module names.

So the `(menhir ...)` stanza has to be in `src/dune` and talk about `language/colisParser` while the `(ocamllex ...)` stanza has to be in `src/language/dune` and talk about `colisLexer`...

benozol:
Sorry `make clean build` still complains:

```
File "src/language/colisLexer.mll", line 3, characters 5-16:
Error: Unbound module ColisParser
```

Did you move the lexer and parser back to src/language on purpose?

Niols:
Pff, I did, but indeed, it does not work. I'll see with the people from Dune. In the meantime, you can revert that last commit…

benozol:
Alright, that one works. Thanks.

Niols:
Sorry about all the noise. I'll fix that when I understand what's going on.

benozol:
No problem. I have merged everything  into `master`. You can try the concrete execution with `make && concrete_colis colis examples/test4.cls` if you want!

PR #8, opened 2018-10-03
Add support for Shell scripts

Niols:
**DO NOT MERGE**

This PR adds support for Shell scripts in the project. It requires a dependency in Morsmall. The conversion is quite complicated. I adapted it from the sourcil-poc project. I need to read it again and to add some documentation. The PR is here in case you want to start reviewing the conversion. @benozol I'm interested by what you think about the conversion of expressions in particular.


src/dune
```diff
@@ -6,5 +6,5 @@
 (executable
  (name concrete_colis)
  (public_name colis)
- (libraries why3.extract)
+ (libraries why3.extract ppx_deriving_yojson.runtime morsmall)
```
benozol: I cannot find morsmall on opam. Is there a release available?

src/language/fromShell.ml
```diff
@@ -1,4 +1,325 @@
...
+
+and word_list__to__list_expression word_list =
+  List.map word__to__string_expression_split_requirement word_list
+  |> List.map
+       (fun (string_expression, split_requirement) ->
+         (string_expression,
+          match split_requirement with
+          | Impossible -> raise (Unsupported "mixed words")
+          | DoesntCare | Split -> C.Split true
+          | NoSplit -> C.Split false))
```
benozol: Nice!

src/language/fromShell.ml
```diff
@@ -1,4 +1,325 @@
...
+       (fun statement assignment' ->
+         let assign = assignment'__to__assign assignment' in
+         C.SSequence (statement, assign))
+       (assignment'__to__assign assignment')
+       assignment'_list
+
+  | Simple (assignment'_list, word' :: word'_list) ->
+     let name = word'__to__name word' in
+     let args = word'_list__to__list_expression word'_list in
+     assert (assignment'_list = []);
```
benozol: I do not know the Morsmall AST at all – but assignments and calls are represented by the same variant `Simple`?

src/language/fromShell.ml
```diff
@@ -1,4 +1,325 @@
...
+(* ============================ [ Expressions ] ============================= *)
+(* We flag each string_expression with its requirements in term of
+   splitting (either Split, NoSplit or DoesntCare). When
+   concatenating, we ensure that the requirements are compatible so
+   that each string_expression has a unique requirement. This is what
+   then gives the 'split' flag in CoLiS. *)
+
+and word_component__to__string_expression_split_requirement = function
+  | Literal s when List.exists (String.contains s) ifs ->
+     (C.ELiteral s, NoSplit)
```
benozol: Hmm… literals that contain an IFS should not be split because they must have had quotes in concrete syntax?

benozol:
I have added a few comments but often I am too unfamiliar with the Morsmall AST. The tests may be pretty important. How should we organize them? #9

src/dune
Niols: No, there is no release. Actually, I'm not sure that Morsmall is really stable right now. I see several possibilities (there might also be others):
- release Morsmall anyway (0.1?) on OPAM and depend on it
- install Morsmall ourselves every time for now (`make && make install` or `opam pin morsmall .`)
- move Morsmall to `src/morsmall` and define a library `symbolic.morsmall` for now. When Morsmall becomes stable / usable for other tools, move it back to its own project (the same goes for constraints, I plan to put them in `src/morsmall` in the library `symbolic.constraints` but move them away as soon as them make a plain coherent block)

What do you think?

src/language/fromShell.ml
Niols: Yes, it comes from the shell, all the following are so-called "simple commands":
- `>/dev/null`
- `echo foo`
- `echo foo >/dev/null`
- `X=a`
- `X=a >/dev/null`
- `X=a echo foo`
- `X=a echo foo >/dev/null`

In fact, in Shell, a "simple command" is composed of three lists (assignments, words and redirections) where one of them is non-empty. In Morsmall, I managed to move redirections away (but actually, it might be a mistake so I might move them back soon) but I couldn't really separate assignments and what we like to think as "calls".

src/language/fromShell.ml
Niols: Actually, single quotes are even lost in Morbig, which I find quite surprising considering that Morbig is supposed to provide a **concrete** syntax tree which, in my opinion, should be in bijection with the textual representation. So you can have:
- `'hello you'` → `Literal "hello you"`
- `hello you` → `Literal "hello"` and `Literal "you"`
- `'hello' 'you'` → `Literal "hello"` and `Literal "you"`

When converting to CoLiS, we don't care about the difference between `hello` and `'hello'` (and even `"hello"`). However, when we get a literal containing a value from the IFS, we can't allow ourselves to have it in a string_expression that will be split afterwards in CoLiS, hence the NoSplit requirement.

I'm not sure if I was really clear here…

Niols:
*This is really WIP*. I had to finish it fast before doing something else today, so I haven't even tested it so far. I still have quite some work on that before I consider it fine. I'll try to achieve that tomorrow morning so that we get a working toolchain Shell → CoLiS → execution. That would really be nice. It would also give us numbers about how much of the corpus we are able to parse with our hypotheses.

src/language/fromShell.ml
Niols: Haha, glad that you're happy ;-)

By the way, now that I read it again, I would rather say that DoesntCare goes with NoSplit to `C.Split false`. Any opinion on the topic?

src/dune
benozol: Maybe integrating morsmall in the git repo as a submodule? That would give us control over the exact version being used, and when it should be upgraded.

Niols: Hum. Say we do that, what would be the steps to follow for the others in order to get the project and build it? Would they have to go in the submodule and pull it and build it? I'm really unfamiliar with this process. If I get it right, that's:
- install Morsmall ourselves every time for now
except we have a nice way to control which *commit* of Morsmall we should use?

src/language/fromShell.ml
Niols: Yes, hence the name of the constructor ^^ But then, we have the choice and we can have a preference (like saying "uh, splitting is dangerous!")

src/language/fromShell.ml
Niols: You mean for the "simple commands"? ;-)

Niols:
It turns out I'm quite happy about that work after all. I have a few things to clean up, but then I think we can merge it.

Niols:
With this last commit, I get on the Debian Sid corpus:

```
Number of scripts: 31968

Different return codes:
    387 Syntax Error
   7550 Unsupported feature
  24031 OK

Non-supported features:
      1 special builtin: exit
      1 special builtin: shift
      2 asynchronous separator &
      2 empty program
      2 mixed words
      2 special builtin: eval
      4 special builtin: exec
      4 special builtin: export
      5 special builtin: unset
      9 special builtin: .
     11 special builtin: continue
     12 special builtin: trap
     13 (word_component_DoubleQuoted)
     14 assignment prefixing simple command
     26 special builtin: break
     36 cd
     44 special builtin: :
     51 here document
     53 (word_component)
     99 (word_to_name)
    115 special builtin: set
    148 other redirections
   1330 function
   1998 absolute source
   3568 case
```

This is the **first** unsupported feature met while translating from Shell. We can however see what our first targets will be: case and source. case wont be a problem because it can be translated to CoLiS (although we have to be careful). source might be a bigger problem.

Anyway, right now, we handle 75% of the scripts in the corpus and I think we can merge this PR (what do you think @benozol?)

benozol:
Great! I’ll try to rebase on master. Can you have a look at #13 ?

src/dune
benozol: Or otherwise `opam pin` ?

Niols: Well that's fine with me, but I wonder about others using the project

PR #20, opened 2018-10-05
Rename statement into instruction and introduce program as a wrapper

### LINKED ISSUE #11: Add a type "program", wrapper around "statement"?
It would be

    type program = statement

in our case (it could be `statement list` but I don't think it adds anything). It would not have a behaviour as an output but only true|false.

It's a small change, but it would make things a bit cleaner imho.

benozol:
Fixes #11

### DIFF
--- a/src/colisParser.mly
+++ b/src/colisParser.mly
@@ -7,7 +7,7 @@ open Syntax__Syntax
 let rec concat = function
   | [] -> assert false (* parsed as a non-empty list *)
   | [se] -> se
-  | se :: ses -> EConcat (se, concat ses)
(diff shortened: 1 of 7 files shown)


Niols:
So 'instruction' is nice as it will avoid the clashes with 'state' but it introduces a clash with 'input'. Which one is best in your opinion? Maybe instruction, because we see 'state' more often than 'input'. I can't find any other name, sadly.

Niols:
Indeed, the common prefix between input and instruction is much shorter than the one between state and statement. The PR looks good to me.

PR #21, opened 2018-10-05
Stdout in state

benozol:
Here is another one.

But I am confused about the correct semantics of pipe. Are changes to the (variable) environment in both branches of a pipe ignored? @Niols could you check the tests in tests/pipe-var.cls and tests/pipe-var.oracle?


benozol:
Here is another one.

But I am confused about the correct semantics of pipe. Are changes to the (variable) environment in both branches of a pipe ignored? @Niols  could you check the tests in `tests/pipe-var.cls` and `tests/pipe-var.oracle`?

Niols:
It's actually unclear to me, looking at the standard. If I remember correctly (I'll have to find a reference for that claim), all the first parts of a pipeline are in a subshell while it's not specified if the last one is.

Niols:
I really can't find it in the standard. The weird part is that the standard insists on the case where you have a not and a subshell:

> If the pipeline begins with the reserved word ! and command1 is a subshell command, the application shall ensure that the ( operator at the beginning of command1 is separated from the ! by one or more <blank> characters. The behavior of the reserved word ! immediately followed by the ( operator is unspecified.

An experiment shows that both bash and dash ignore the changes of the environment in all the parts of the pipeline.

benozol:
> An experiment shows that both bash and dash ignore the changes of the environment in all the parts of the pipeline.

Yes that’s also what I observed in `dash` and `bash --posix` and implemented. So let’s see later if any other shells disagree?

By the way, the fact that the parts of the pipeline run in subshells also influences the possibility to translate pipes of sequences into Shell, for example:

```
pipe begin foo; bar end into baz; qux epip
```
into
```
(foo; bar) | (baz; qux)
```
Or do you imagine a translation without subshells?

Niols:
Hum, I don't understand what you mean about the translation? Do you mean that

    (foo; bar) | (baz; qux)

and 

    { foo; bar; } | { bar; qux; }

are equivalent in that situation? Then I guess we can translate to whichever of these two expressions

benozol:
Alright, curly braces! I wasn’t aware that they can be used outside function definition :o

benozol:
We are almost there (all tests pass). But I cannot find out why the CI fails – or more precisely, why the library `ppx_protocol_conv` is not installed for testing:

https://travis-ci.org/colis-anr/colis-language/jobs/439602669#L695

@Niols do you have an idea?

tests/for-false.sh
```diff
@@ -1,4 +1,4 @@
-for x in ''; do
+for x in 'x'; do
```
Niols: Hum, why this change?

Niols:
Yes, I see the problem with the dependency. I got an e-mail saying "Explicit protocol conv library to dune" but I can't find it on GitHub and I don't know what it's refering to. The problem comes from the Dockerfile. Do you want me to fix it in this branch or after the merge?

Niols:
It works, but the targets 'install' and 'uninstall' are not in the Makefile. So either we remove them from the Dockerfile or we add them to Makefile:

    install:
        dune install

    uninstall:
        dune uninstall

Niols:
And that should (hopefully) be the last thing we need before the merge (pfiouh)

benozol:
OK let’s see :crossed_fingers: :crossed_swords:

Niols:
It passed for 4.05. Except unexpected stuff, it should be fine with the other versions (btw, is 4.05 really the minimal version? I'll try to lower that, but not in this PR)

Niols:
Yay \o

benozol:
Ah that’s nice. Now we can start taking tests seriously ;)

BTW `tests/str-subshell` was really a journey through many murky corners of shell semantics…

PR #22, opened 2018-10-05
Compare tests with oracle using the `diff` utility

benozol:
Preparation to remove the functionality from the colis executable

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ build: extract-why3
 
 test: build
 	dune runtest
+	./run_tests.sh
 
 clean: clean-why3
 	dune clean
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,6 @@ build: extract-why3
 
 test: build
 	dune runtest
-	./run_tests.sh
 
 clean: clean-why3
 	dune clean
--- a/run_tests.sh
+++ b/run_tests.sh
@@ -1,13 +1,69 @@
-#!/bin/sh
+#!/bin/bash
+set -e
+
+COLIS="./concrete_colis"
 
 allgood="yes"
-for f in tests/*.cls; do
-    if ./concrete_colis -compare -colis "$f" > /dev/null 2>&1; then
-        echo "SUCCESS $(basename "$f" .cls)."
-    else
-        echo "FAILURE $(basename "$f" .cls)." >&2
-        allgood="no"
+
+error() {
+    file=$1
+    shift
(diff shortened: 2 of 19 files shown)


Makefile
```diff
@@ -6,6 +6,7 @@ build: extract-why3
 
 test: build
 	dune runtest
+	./run_tests.sh
```
Niols: Actually, this should go in a Dune file rather than in the Makefile. But I'll take care of that eventually

run_tests.sh
```diff
@@ -1,13 +1,69 @@
-#!/bin/sh
+#!/bin/bash
```
Niols: Aw, sad :-(

Niols: Where do we need bash stuff?

run_tests.sh
```diff
@@ -1,13 +1,69 @@
...
+    base=$(basename "$file" $extension)
+
+    # Derive oracle file name
+    oracle="tests/$base.oracle"
+    if ! [ -e "$oracle" ]; then
+        error "$file" "Missing oracle $oracle"
+        continue
     fi
+
+    outbase=$(mktemp -u -t "$base-XXXXX")
```
Niols: I've had problems before with `mktemp` not being available on all systems. But let's keep it that way for now, I'd say

Makefile
benozol: OK I’ll remove it from the PR

run_tests.sh
benozol: Oh forgot to change it back, will do

run_tests.sh
benozol: Is `tempfile` more generally available?

Makefile
Niols: I'm not sure you should. Maybe it should be the job of the future PR that'll pass tests under Dune?

benozol:
And I was thinking to replace `RESULT: true|false` into `RETURN: <N>` in the oracle files. This would allow for oracle files that detect syntax errors or errors in the conversion. OK?

Niols:
Yes, it's nice to have the integer. Where should we document which integer goes with which behaviour of the colis command?

benozol:
What about describing the return codes in the README and/or the future man-page?

I left my computer and will probably not be able to review your other PR on the CLI before Monday :(

Bon weekend !

Am 5. Oktober 2018 16:06:14 MESZ schrieb Nicolas Jeannerod <notifications@github.com>:

Issue #12, opened 2018-10-04
Command Line Interface


Niols:
Apparently, for now, we want to put everything in the same executable. This is fine to me. However, as it will have to do a lot of things, the command line will be slightly complicated and deserves that we think a bit about it.

What I would like to be able to do with the tool:
- handle colis and shell inputs the same way (this is already what you added with `-shell` and `-colis`)
- be able to only parse (I don't know about colis scripts, but it would be useful to measure how many shell scripts are accepted by our tool without executing anything)
- be able to print colis or shell. This would allow us to see what the translation/s does/do
- be able to execute in a concrete or symbolic way

Question:
- do we want to handle several files at once? I'm not sure if it's really useful (we can always have a wrapper around the tool) and it makes things a bit more complicated.

Maybe:
- `--colis` (to specify the format of the input)
- `--shell` (the default?)
- `--print-colis` (on standard output?)
- `--print-shell` (or maybe just `--print` and we use the selector `--colis`|`--shell` to choose the other one? But then, maybe print is a bad name)
- `--concrete` (the default, I'd say? Well, for now, we don't have much choice)
- `--symbolic`

Niols:
- `--realworld`? to be plugged to utilities found in the PATH

benozol:
> --realworld? to be plugged to utilities found in the PATH

Agreed, this would be useful.

> --print-shell (or maybe just --print and we use the selector --colis|--shell to choose the other one? But then, maybe print is a bad name)

Two printing options matching the parsing options are fine I think.

---

What do you think of the following command line syntax? Defaults are marked with asterisks.
```
./colis [--help | [--run* [--realworld] [--compare[=ORACLE]] | --run-symbolic | --print-colis | --print-shell] [--shell* | --colis] FILE ]
   --run                Concrete execution
   --run-symbolic       Symbolic execution (not yet implemented)
   --print-colis        Print the CoLiS script
   --print-shell        Print the Shell script
   --realword           Use system utilities in concrete execution
   --compare[=ORACLE]	Execute of FILE and compare results from file ORACLE (or from file "$(basename FILE .cls .sh).oracle")   
   --shell              Use the Shell parser
   --colis              Use the CoLiS parser

```

We could also use subcommands (like `colis concrete`, `colis print`, etc) but I am not sure if this makes the CLI really easier.

Niols:
I like this command line. Maybe the following comments:
- do you really want to avoid having a default between `--shell` and `--colis`?
[...]

Issue #6, opened 2018-10-02
Test suite and continous integration


claudemarche:
The former test suite from the Gforge repository should be reused, for replaying tests manually and also for continous integration.

* import the former `.cls` files and adapt their concrete syntax
  though, we can forget about the tests that involve `date` or commands that modify the filesystem: `mkdir` and such
* import the corresponding `.oracle` files as is
* implement in OCaml a check of conformance between the result of concrete interpreter and the contents of oracle files
* add a continous integration process that checks the conformance of all tests with their oracles

Niols:
Linked to #9. I can take care of the continuous integration part as I'm used to setting that kind of things up (with Docker + Travis for instance)

Niols:
Having Docker is actually much more useful than for testing only. Especially considering that Why3 is not that common to have on one's machine and not that easy to setup in the right conditions

Issue #9, opened 2018-10-03
Organize tests


benozol:
The branch `fully-imperative-concrete` contains a directory `tests` with colis scripts and oracle files, and a shell script `run_tests.sh` to compare the execution of scripts with their oracle, for all `tests/*.cls`.

The branch `from-colis` uses files under `examples/shell/` for testing.

Should we create a single command to run the parsing tests and the execution tests?

Niols:
There is already a single command to run tests, which is `dune runtest`. We can also have `make test` that just runs that. Then, we can put tests at several places in the repository (including inline tests in the code) and they will all be detected an run by dune (provided we give it the right `dune` files).

There might be a difference between examples and tests when tests are here for precise tests cases and examples are here to show how to tool is supposed to work. However, for now, I guess that everything can go in `tests/`? But maybe `tests/shell`, `tests/colis`, `tests/...`

Also, what do you think of having the `test` command using Why3 to check the proofs? Then, maybe not `make tests` but `make checks`? Or maybe we could have `make tests` for "regular" tests, `make check-proofs` for Why3 proofs and `make checks` that depends on them both?

Issue #7, opened 2018-10-03
Integrate Morsmall parser


benozol:
The shell parser can be activated with command `./concrete_colis -shell FILENAME` but the implementation is actually missing.

The functions `{FromColis, FromShell}.parse` currently have a signature `filename:string -> statement`. Maybe we should change it to a more general argument type that works with files and strings of code.

Niols:
I'm already on it. PR coming soon (this afternoon). However, we will then require Morsmall


-------- Message d’origine --------
De : Benedikt Becker <notifications@github.com>
Envoyé : 3 octobre 2018 10:51:15 GMT+02:00
À : colis-anr/symbolic <symbolic@noreply.github.com>
Cc : Nicolas Jeannerod <niols@niols.fr>, Assign <assign@noreply.github.com>
Objet : [colis-anr/symbolic] Integrate Moresmall parser (#7)

The shell parser can be activated with command `./concrete_colis -shell FILENAME` but the implementation is actually missing.

The functions `{FromColis, FromShell}.parse` currently have a signature `filename:string -> statement`. Maybe we should change it to a more general argument type that works with files and strings of code.

Niols:
See #8

Niols:
I've read about submodules and it might be a way to do what we want. Except that the commands are a bit painful:
- Cloning is now: `git clone --recursive`
- Pulling when the version of the submodule has changed should be `git pull && git submodule sync --recursive && git submodule update --init --recursive` and sadly, there is no easier way to do it

However, Dune has no problem with that and is perfectly fine with building the two required libraries.

Niols:
@benozol I can do that right now if you still think it's the best option. I think it's OK. Not the best, but probably the least bad in order to satisfy everyone.

I'd probably add a directory "submodules" (or "external"?) at the root so that they don't appear in the source tree, which might be weird (what do you think about that?).

Compilation will work nicely with submodules that compile with Dune. However, you wanted to add why3-ide as a submodule too, right? I don't really know how to achieve that in a clean way

Niols:
After the merge of #36 into master, I'll just copy Morsmall from its git repo to `lib/morsmall` for now. I think it's the simplest solution. And if it is stable enough in the end, I'll just move it back to its repository.

benozol:
[...]

PR #23, opened 2018-10-05
Command line interface, tests and continuous integration

Niols:
This is an implementation of what have been discussed in #12. What do you think about it? I also wanted to change the tests to use the new oracle, but it turns out you did it in #22. There are still a few changes required like using `bin/colis` un `run_tests.sh`. We can take care of these changes either in #22 or in here.

Also, I'm not sure about everything here, I'd gladly welcome your review @benozol.


src/colis.ml
```diff
@@ -0,0 +1,97 @@
...
+       | Shell -> FromShell.parse
+  in
+  match get_action () with
+  | Run ->
+     (
+       let open Interpreter in
+       (* FIXME: realworld *)
+       let state = empty_state () in
+       interp_program empty_input state program;
+       print_string (!(state.stdout) |> List.rev |> String.concat "\n");
```
Niols: This seems to be a problem because none of the tests passes. But I thought that the first element of `!(state.stdout)` (that is, the last of `List.rev !(state.stdout)`) was the current line we were writing to?

Niols:
Alright. I added a Dockerfile to build the tool and I used it to setup continuous integration. It seems to work fine now. A few notes:
- The CI test fails because the tests do, not because the CI is broken. About the tests, see src/colis.ml, line 71. We should fix that before merging the PR.
- This PR now closes #6, I'd say, and possibly #9, now that the point about replaying proofs is in #24.
- This PR still closes #12, I guess.
- What we decide about submodules or not in #7 can change what happens in the Dockerfile about building Morsmall (as, for now, we just clone Morsmall's repo in the Dockerfile and install by hand, which is not very robust)

Niols:
While we are at it, there is something else that I would like to add to the command line interface: I think that we should be able to send command-line arguments to the interpreter. Either by saying that all extra arguments of the command line are stacked in a list that is given to `inter_program` as its args. An other possibility is to wait for `--` and to use only arguments that are after this token, which could make the command line a bit clearer for a user. I'll implement the former as it's really easy, but what do you think about it @benozol?

Niols:
Here you go, waiting for your review. Bon week-end! :-)

Niols:
There are still two things I would like to do:
- move shell scripts in `examples/shell` to `tests/good` and use them as tests
- maybe rewrite the test runner in OCaml?

src/colis.ml
benozol: Is this still an issue?

src/colis.ml
```diff
@@ -0,0 +1,100 @@
...
+    match !file with
+    | None -> file := Some new_file_or_arg
+    | Some _ -> args := new_file_or_arg :: !args)
+
+let speclist =
+  [ "--run",          Arg.Unit (set_action Run),         " Concrete execution (default)";
+    "--run-symbolic", Arg.Unit (set_action RunSymbolic), " Symbolic execution";
+    "--print-colis",  Arg.Unit (set_action PrintColis),  " Print the CoLiS script";
+    "--print-shell",  Arg.Unit (set_action PrintShell),  " Print the Shell script";
+    "--realworld",    Arg.Set realworld,                 " Use system utilities in concrete execution";
```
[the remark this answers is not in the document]
Niols: It would probably make a lot of sense. I don't know how I would do that in a clean way, though

src/colis.ml
Niols: I wonder now. Maybe it was an issue in the way the Shell test runner was comparing the outputs?

src/dune
```diff
@@ -4,7 +4,7 @@
 (menhir (flags --explain) (modules colisParser))
 
 (executable
- (name concrete_colis)
+ (name colis)
  (public_name colis)
- (libraries zarith str ppx_deriving_yojson.runtime morsmall)
+ (libraries why3.extract str ppx_deriving_yojson.runtime morsmall)
```
benozol: Sorry you were right initially: `why3.extract` is not required any more, all why3-libraries should be extracted and this should be `zarith`.

Niols: Oh, OK, I'll fix that

src/colis.ml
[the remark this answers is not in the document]
benozol: > … how I would do that in a clean way

Probably not with `Arg`, but you can just check before executing the actions. I am unsure how `colis --run-symbolic --realword --run FILE` should behave…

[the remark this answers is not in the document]
Niols: I currently do not authorise to use more than one of `--run`, `--run-symbolic`, `--print-colis` and `--print-shell`. But yes, I should just add a check after the arguments parsing…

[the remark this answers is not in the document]
Niols: OK, I have just added a test after the call to `Arg.parse`. I'm not sure I like how I parsed the arguments though…

tests/run_tests.ml
```diff
@@ -0,0 +1,107 @@
+let (||>) f g x = f x |> g
```
benozol: A few remarks:

1. What is the advantage of rewriting the Shell script in OCaml (Shell is, after all, a wonderful language for running, piping and testing a few command :smiling_imp:)
2. Oracles should be easy to read and write, and I have the impression that JSON is more targeted to simplify the reading and writing for computers. If you like standard formats, would YAML be an alternative? (PS: I saw that you the executable `run_tests` prints errors nicely so this may only be relevant as long as Oracles have to be *written* manually.)
3. Oracles should be explicit, e.g., `{}` should be `{'return_code': 0; 'output': ''}`. WDYT?
4. Shouldn’t standard output be called `stdout` instead of `output`, in case we add standard error one day?
5. The description of the tests and oracle format in `README.md` should be updated.

Niols: 1. It's a good question and I'm not entirely convinced myself. I like Shell scripting precisely for that kind of small scripts. In that case, OCaml also requires us to write administrative code like getting everything from an `in_channel` into a string. However, once it is done, the tests runner should be more robust. In particular, it allows for a better handling of things like the list of the arguments (idk how I would do it in a clean way in Shell, we'd have to use xargs or something). The fact that we use a serialisation format for the oracles means that we can extend them more easily and that we can support more expressive oracles. Long story short, I see several small advantages in using OCaml instead of Shell, but I don't mind if we decide to stick with Shell.

2. JSon is not that bad to read imho but Yaml would be better for humans. However, I also chose JSon because `ppx_deriving_yojson` allows for direct parsing from JSon to OCaml types and because it was already a dependency (because of Morbig). I don't understand the comment about errors printing in `run_tests`?

3. Agreed, for the outputs (that is, return_code and stdout). The inputs (like the list of arguments) could remain implicit, or should they be explicit too?

4. Yup, agreed.

5. I'll do that after we decide to keep this modification or not

benozol: 1. Agreed wrt extensibility: Parsing oracle files in shell will become painful at the latest when we want to include stdin and expected stdin after execution
2. Re comment about errors printing: Your implementation displays stdout nicely, which at least invalidates the argument of bad readability of JSON. Writing the stdout in a JSON string is still quite fiddly
3. Yes these are a sane defaults I guess
5. OK

Niols: About 2., I think that one can write actual line-returns instead of "\n":

    { "stdout": "a
    b c
    d
    ",
      "return_code": 2 }

but I'm not sure if it's really convincing.

Niols: Or we could provide files containing the expected output? Like:

    type stuff = Raw of string | File of string
    type oracle =
      { stdout : stuff ;
        return_code : int }

and one would have in `test.oracle`:

    { "stdout": ["File", "test.out"],
      "return_code": 2 }

and in `test.out`:

    a
    b c
    

but I'm not sure it's great as it forces us to read more files. Also, I like the serialisation format as one can clearly see whether there are spaces at the end of the lines, or a final newline, etc.

Niols: There is an OCaml serialiser and deserialiser for several datatypes: https://github.com/andersfugmann/ppx_protocol_conv. We could try to use that, if you prefer Yaml?

Niols:
By the way, there are three tests that don't pass:
- `false-while.cls` and `false-while.sh` because the while loop is not supported by the interpreter (if I'm not mistaken)
- `for-false.sh` because there seems to be a problem in Morsmall (we get an assert failure)

We should fix them soon, because in the current situation, we would probably miss the introduction of a regression.

tests/run_tests.ml
benozol: > provide files containing the expected output … forces us to read more files

I agree, oracles should be concise and not distributed over several files

YAML is a nice human/machine compromise I think but we would still have to indent the stdout like this:

```yaml
stdout: |
  a
  b c
  d
  
return_code: 2
```

Another possibility would be a custom parser for a simple text format, e.g.:
```
ARG: arg1
ARG: arg2
RETURN: 2
STDOUT:
a
b c

```

But I am not sure if anything is entirely convincing, so I agree with JSON if you see most advantages there (*in dubio pro author*?)

benozol:
> the while loop is not supported by the interpreter

I am working on it!

tests/run_tests.ml
Niols: I agree, it's better than JSon in term of human-readbility. The only thing that worries me is the handling of final newlines. How would one make the difference between `"a\nb c\nd\n"` and `a\nb c\nd"` in your example?

benozol: Either using a trailing empty line (as in the example above), or by an explicit delimiter (like in JSON :crying_cat_face:  ):
```
STDOUT BEGIN:
a
b c
d

END

benozol:
BTW oracle files (`*.meta`) are missing for a few tests in this branch (`command-not-found`, `pipe-var`, `grep-none`). Do you have them somewhere or should I create and push them here?

tests/run_tests.ml
Niols: Urgh.

benozol:
> BTW oracle files (*.meta) are missing for a few tests in this branch (command-not-found, pipe-var, grep-none). Do you have them somewhere or should I create and push them here?

I have created and added them to #21 (which is rebased on #23). You can cherry-pick from there or we wait for the big merge. Only `for-false.sh` is still failing.

tests/run_tests.ml
Niols: I gave a try to meta files in Yaml. It adds one dependency (in Yaml and ppx_protocol_conv) but only for testing, so I guess it's OK. I had one issue with the Yaml module in OCaml, we'll see if they solve it. 

Anyway, I agree on the fact that .meta files are now much more readable (and I even kept outputs as encoded strings, we should use the other syntax you showed me).

We do you think about it? Personally, I'd rather add a dependency and stick with a known format than write a parser for our custom format, but if you really prefer something custom, we can go for it.

Niols:
Hey. I just changed .meta files to the Yaml format. If you like it, I just have the following to do and we can go for the big merge:
- rename the "output" field into "stdout"
- add fields for "stdin" and "stderr" (while I'm at it)
- change the meta files to use nice strings instead of the encoded format (can you give me an example?)

By the way, I was wondering if we should specify what is the input and the output in the meta file? Like:

    input:
      arguments: ...
      stdin: ...
    output:
      stdout: ...
      return_code: ...

rather than:

    arguments: ...
    stdin: ...
    stdout: ...
    return_code: ...

What do you think about that?

Niols:
Alright. What do you think about the current state of the PR? We can still take our time to make a few things better, but I'm under the impression that it's rather good. All the tests that don't pass in this PR should pass after the merge (you fixed two of them, I fixed one in Morsmall).

benozol:
Very nice indeed! Let’s merge (I can do).

PR #28, opened 2018-10-11
Concrete interpreter soundness

### LINKED ISSUE #24: A full-check target to the Makefile that replays the proofs
Once we add proofs, there should be a way in the Makefile to call `why3 replay` with the right arguments. Or do you think it should belong to the `test` target? (in which case, we really have to rename that target into `check` ^^)

### LINKED ISSUE #26: Proof soundness of concrete interpretation
And see #24

benozol:
Resolves #26 and fixes #24


Makefile
```diff
@@ -1,4 +1,4 @@
-.PHONY: build test clean extract-why3 clean-why3
+.PHONY: build tessemanticst clean extract-why3 clean-why3
```
Niols: Uh?

Makefile
```diff
@@ -7,10 +7,23 @@ build: extract-why3
 test: build
 	dune runtest
 
+check: check-concrete-semantics check-concrete-interpreter
```
Niols: So we keep two different targets, test and check?

src/concrete/interpreter/why3session.xml.bak
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
Niols: Do we really want to include the `.bak` files?

Makefile
benozol: Ups.

Makefile
benozol: New proposition: one target `sure` that combines the targets `test` and `proof` (or `prove`? not sure if this should be grammatically a noun or a verb…)

src/concrete/interpreter/why3session.xml.bak
benozol: That is indeed not very useful.

Niols:
Does pushing the shapes bring something to the user / the replayer of proofs?

benozol:
Good question. Would it allow replaying only VCs with obsolete proofs (using `wh3 replay --obsolete-only`)? Or don’t we care and replay all proofs at every run of the CI?

Niols:
Well, I don't know, what are they for?

Niols:
I mean in general in Why3

benozol:
> Does pushing the shapes bring something to the user / the replayer of proofs?

Yes, they are required for Why3 to pair the VCs of the program with previous proofs.

.travis.yml
```diff
@@ -7,12 +7,19 @@ services:
 ## are in fact tags for the ocaml/opam2 Docker image that we pull.
 
 env:
-  - TAG=4.05
-  - TAG=4.06
-  - TAG=4.07
-
-## Two steps: we first build the Dockerfile. If it succeeds, we try
-## running tests, installing and uninstalling in the Dockerfile.
+  global:
```
benozol: Opam packages are pinned from here…

benozol:
Hi @Niols What do you think? Can I merge this PR?

.travis.yml
```diff
@@ -7,12 +7,15 @@ services:
...
 
 env:
-  - TAG=4.05
-  - TAG=4.06
-  - TAG=4.07
+  global:
+    - "WHY3COMMIT=#41d0c60df"
+  matrix:
+    # - TAG=4.05
+    # - TAG=4.06
```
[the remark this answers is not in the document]
Niols: Ow, that's bad. And my fault. Although I don't mind loosing older versions of OCaml, I think that forcing 4.07 is a bit too strong. I'll fix that on master after the merge.

Dockerfile
```diff
@@ -1,33 +1,78 @@
-## =============================== [ Config ] =============================== ##
+## =============================== [ Basis ] =============================== ##
```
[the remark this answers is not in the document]
Niols: You indeed added quite a lot. I was trying to avoid using specific tools like `apt` in order to have tests working on different distributions, but I admit it might become quite hard now.

Also, is it possible to add the pinning as a dependency in the `.opam` file? I think it should be but I have no idea how. Also, it would be great to be able to have the provers as external dependencies of this project, but it only works if there are packages for them (and I don't think there is?)

Niols:
I'm trying to replay all that on my machine. The opam file was incomplete, I'll push a fix. Also, the default target of the Makefile is `make ci`. Is it on purpose? I'd rather have `build` as the default target. Especially considering that the CI dependencies are much heavier than the build dependencies (because of all the provers -- one does not necessarily want to install them all).

benozol:
> I'd rather have build as the default target. 

Of course!

Dockerfile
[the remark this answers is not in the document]
benozol: Do you think there is a way to install `wget` and the compilation dependencies of `why3` (`autoconf` and `automake`) without `apt`?

AFAIK most provers are unfortunately not packaged (on Debian only Alt-ergo but in an old version).

Niols:
OK, I made a few little modifications. If the CI is happy, then I am!

Dockerfile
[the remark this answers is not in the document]
Niols: > Do you think there is a way to install `wget` and the compilation dependencies of `why3` (`autoconf` and `automake`) without `apt`?

Sadly, except with a small script that tests the package manager, I don't think so. So APT will be just fine.

> I have just discovered the field `pin-depends`

That's exactly what I was looking for, but I couldn't find it in the documentation (it's not that easy to find stuff in there)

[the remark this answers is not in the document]
benozol: Not sure if this will be as easy as adding some `pin-depends` to the opam file.

The dependencies are installed in the container using `opam depext -i $(opam show . -f depends: | cut -d '"' -f 2)`, so the information about `pin-depends` wouldn’t be used here (but breaks the script by adding `Would pin the following: …` to the output of `opam show`). We would instead need to pin the dependencies before, but translating the output of `opam show . -f pin-depends` to `opam pin` instructions is little fun.

Can’t we just do `opam install .` to ensure the dependencies and `make ci` later?

```
pin-depends: [
  [ "morsmall.git" "git+https://github.com:colis-anr/morsmall" ]
  [ "why3.git"     "git+https://gitlab.inria.fr:why3/why3.git#41d0c60df" ]
]
```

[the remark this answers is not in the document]
benozol: > trying to avoid using specific tools like apt

I have the impression that this is what opam’s `depexts` and `conf-*` dependencies are good for. We could use it for `wget` but Why3 should depend on `autoconf` and `automake`. Does this make sense?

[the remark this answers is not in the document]
Niols: > The dependencies are installed in the container using `opam depext -i $(opam show . -f depends: | cut -d '"' -f 2)`,

This little trick was useful to avoid using `apt` (or any specific package manager). The best solution would be for `opam depext` to support `opam depext . --deps-only` just like `opam install`. An other solution in our case is to simply install the dependencies with `apt` and then to use `opam install . --deps-only`.

> Can’t we just do `opam install .` to ensure the dependencies and `make ci` later?

Yes, we can! Also, are the names supposed to be "morsmall.git" and "why3.git" or just "morsmall" and "why3?

> I have the impression that this is what opam’s `depexts` and `conf-*` dependencies are good for. We could use it for `wget` but Why3 should depend on `autoconf` and `automake`. Does this make sense?

Well, Why3 doesn't depend on `autoconf` and `automake` in any way so I don't think this is where the depext should be. However, the dependencies in the prover should go in the depext field (and probably be recommendations only). Except that there are no packages for these.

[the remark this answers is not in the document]
Niols: Also, if you wanted a magic trick:

```sh
opam show . -f pin-depends: 2>/dev/null | tail -n -2 | tr -s '[]"' ' ' | xargs -n2 opam pin -n
```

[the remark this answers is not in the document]
Niols: Actually, tail -n -2 doesn't work that well… So maybe:

```sh
opam show . -f pin-depends: 2>/dev/null | grep -A 50000 '^$' | tr -s '[]"' ' ' | xargs -n2 opam pin -n
```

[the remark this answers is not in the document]
Niols: (let's not have that in our Dockerfile, hm? I really love Shell but it's way too dangerous)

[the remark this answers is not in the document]
Niols: By the way (and this is an important remark), why do you use `:` and not `/`: in `git+https://github.com:colis-anr/morsmall`. opam does not seem to be happy with that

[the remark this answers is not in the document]
benozol: Honestly, I did not find a way yet to make opam happy with my urls for `pin-depends` … any clue?

Dockerfile
```diff
@@ -7,21 +7,33 @@ FROM $IMAGE
...
-RUN sudo apt-get update && sudo apt-get install -y wget autoconf automake
+WORKDIR /home/opam/colis-language
+COPY . .
+RUN sudo chown -R opam .
+
+# The following could be accomplished by something like `opam install --only-deps` if it was supported...
 
-RUN [ -n "$MORSMALLPIN" ] && opam pin -n "$MORSMALLPIN"
-RUN [ -n "$WHY3PIN" ] && opam pin -n "$WHY3PIN"
+RUN sudo apt-get update && sudo apt-get install -yy -qq wget autoconf automake
```
Niols: This line could go before the `COPY . .`, so that it stays cached even though we change stuff in the project files.

Dockerfile
```diff
@@ -7,21 +7,33 @@ FROM $IMAGE
...
 
-COPY *.opam .
-RUN opam depext -i $(opam show . -f depends: | cut -d '"' -f 2)
+# Extract pin-depends from opam file and pin them
+RUN opam show . -f pin-depends: 2>/dev/null \
+  | grep -A 50000 '^$' \
+  | tr -s '[]"' ' ' \
+  | xargs -n2 opam pin -n
```
Niols: So you went for this after all? ;-)

Dockerfile
```diff
@@ -7,21 +7,33 @@ FROM $IMAGE
...
+
+# Extract dependencies from opam file and install their dependencies
+RUN opam show . -f depends: \
+  | grep -A 50000 '^$' \
+  | cut -d '"' -f 2 \
+  > /tmp/dependencies; \
+  opam depext --install $(cat /tmp/dependencies)
```
Niols: Hum, this could be `| xargs opam depext --install` I think?

colis-language.opam
```diff
@@ -15,14 +15,24 @@ authors: [
...
+
 maintainer: "Nicolas Jeannerod <nicolas.jeannerod@irif.fr>"
 
+pin-depends: [
+  [ "morsmall.dev" "https://github.com/colis-anr/morsmall.git" ]
+  [ "why3.dev"     "https://gitlab.inria.fr/why3/why3.git#41d0c60df" ]
+]
+
+depexts: [
+  ["wget" "autoconf" "automake"] {os-distribution = "ubuntu" | os-distribution = "debian"}
```
Niols: These package names also work for Archlinux (I just checked)

colis-language.opam
```diff
@@ -15,14 +15,24 @@ authors: [
...
-  "morsmall"
-  "ocaml"     {build & >= "4.05"}
-  "odoc"      {with-doc}
-  "ppx_protocol_conv_yaml"  {with-test}
   "zarith"
-  "why3"      {build | with-test}
+  "ocaml"                  {build & >= "4.05"}
+  "camlzip"                {build | with-test}
+  "morsmall"               {= "dev"}
+  "why3"                   {= "dev" & build | with-test}
```
Niols: What are the priorities between `&` and `|`? Because I'd say that this means `(= "dev" & build) | with-test` while we want `= "dev" & (build | with-test)`

Niols:
Hey, I made a few remarks, but apart from them, I think it's nice. You can merge whenever you want :-)

Dockerfile
benozol: Well I thought for a project like like this it’s quite genuine to put some serious shell foo

Issue #19, opened 2018-10-04
Should we add positions to the AST? Where?


Niols:

Niols:
I do still believe it would be nice to have positions to allow for better error messages. The positions are present in Morsmall (and used in conversion errors). Adding them to Colis, however, can be annoying.

PR #30, opened 2018-10-16
Split `src/colis.ml` into a library and an executable

Niols:
- It makes the separation of code cleaner
- It allows for an easy use of internal functions (eg for testing purpose)
- It allows for a better documentation of internal functions
- It will allow an other tool to use the internal library directly instead of using the executable


src/colis_cmd.ml
```diff
@@ -0,0 +1,115 @@
...
+      )
+    with
+    | Colis.ColisLexer.LexerError s ->
+       eprintf "Lexing error: %s@." s;
+       exit 2
+    | Colis.ColisParser.Error ->
+       eprintf "Parsing error@.";
+       exit 2
+    | Morsmall.SyntaxError _pos ->
+       eprintf "Syntax error@.";
```
benozol: What about showing the same error message for syntax errors in Shell (`Morsmall.SyntaxError`) and Colis (`Colis.ColisParser`)? (BTW it would be nice if we could have a `pos` for the Colis parser as well and display it. I’ll have a look later…)

benozol:
Very useful PR – I was already alienated by how much logic should be put into the `colis.ml`.

src/colis_cmd.ml
Niols: I agree. If fact, in the lib `Colis`, we could have only one exception for all lexing/parsing errors, what do you think? And indeed, it would be nice to carry a position in the parser (which reminds me of #19: do we want positions in the AST?)

benozol:
OK looks good to me!

Niols:
Do you think we should merge it?

benozol:
Definitely

PR #31, opened 2018-10-17
Symbolic execution

benozol:
For now I have just added a few interfaces from the `ftwu` repo under `src/constraints`. @Niols if you add ftwu to this repo I’ll rebase this branch

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,11 +20,11 @@ extract-why3:
 	rm -f src/why3/*
 	why3 extract --modular --recursive \
 		-D ocaml64 \
-		-D src/language/driver.drv -D src/concrete/driver.drv \
-		-L src/language -L src/concrete \
+		-D src/language/driver.drv -D src/concrete/driver.drv -D src/symbolic/driver.drv \
+		-L src/language -L src/concrete -L src/symbolic \
 		-o src/why3 \
-		map.Const \
-		interpreter.Interpreter
+		interpreter.Interpreter \
+		symbexec.Interpreter
 
 clean-why3:
 	rm -rf src/why3
--- a/Makefile
+++ b/Makefile
@@ -24,7 +24,8 @@ extract-why3:
 		-L src/language -L src/concrete -L src/symbolic \
 		-o src/why3 \
 		interpreter.Interpreter \
-		symbexec.Interpreter
+		symbexec.Interpreter \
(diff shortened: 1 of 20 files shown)


Niols:
We probably should. I was thinking about having a submodule for it, but it will change really fast, so it might not be the best idea. I'd just copy the directory inside `src/ftwu`, WDYT?

benozol:
Yes, that might be a viable approach (also for morsmall): to keep the development of the actually independent libraries in the `colis-language` repo while things are moving fast, and separate and release when they become stable.

benozol:
At least it typechecks currently. @Niols what do you think of the usage of ftwu?

https://github.com/colis-anr/colis-language/blob/577ed67204a1458ddce73a0b4dfca93c8b8d69cd/src/symbolic/symbolicBuiltin.ml#L120-L139

benozol:
> where are the functions exists, and dir defined? Where is the monadic bind defined? Why not in Clause?

Currently only by ad hoc definitions in `Symbolic_builtin` – awaiting the import of `ftwu`.

> maybe we should add val exists2 : ?hint1:string -> ?hint2:string -> (Variable.t -> Variable.t -> t -> disj) -> t -> disj as a helper? And maybe exists3, etc. if we need them?

Not quite sure if this will be necessary

> do you want me to suggest a new interface where everything is in t -> disj? In this PR? In a PR going to branch start-symbolic?

That would be great, in a branch of you choice

Niols:
> > maybe we should add val exists2 : ?hint1:string -> ?hint2:string -> (Variable.t -> Variable.t -> t -> disj) -> t -> disj as a helper? And maybe exists3, etc. if we need them?
> 
> Not quite sure if this will be necessary

OK, I just wondered if it could make the code a bit clearer by allowing multiple quantifications at the same line;

> > where are the functions exists, and dir defined? Where is the monadic bind defined? Why not in Clause?
> 
> Currently only by ad hoc definitions in `Symbolic_builtin` – awaiting the import of `ftwu`.
> 
> > do you want me to suggest a new interface where everything is in t -> disj? In this PR? In a PR going to branch start-symbolic?
> 
> That would be great, in a branch of you choice

OK, I'll go for it then

Niols:
Oh. This is considered merged now. I got kinda lost in all the branches, but I guess #31 was rebased in #36?

Issue #27, opened 2018-10-11
Add functions to the CoLiS language


benozol:
Functions should be added to the CoLiS language to define macros in the specification of shell commands. (By constrast, functions in the Debian corpus can generally be inlined.)

There are two possibilities for adding functions:
1. Function definitions precede the instruction of a `program`; a function environment is added in the evaluation context; and function are called by their name (“the usual way”). E.g.:
```mlw
type instruction = …
| TFunctionCall string list_expression
type function_definition = string * instruction
type program = list function_definition * instruction
```
2. Function definitions are inlined into the call site, which is easy because the shell parser already creates a function environment. E.g.:
```mlw
type instruction = …
| TFunctionCall instruction list_expression
```

Evaluation of 1. allows for defining recursive functions (which is at least necessary for the specification macro of `rm -rf`) but discards the information about function bindings that was already generated by the shell parser. Evaluation in 2. is trivial but prevents the definition of for recursive functions, .

benozol:
Given that the main motivation to add functions is the definition of command macros – which are potentially recursive – we should go with 1.

benozol:
Function definitions in the instructions would allow for local functions. Do local Shell functions conform to POSIX? Are there occurrences in the Debian corpus? Maybe it wouldn’t complicate the semantics either… Where do you see benefits?

Niols:
I wasn't thinking about local functions. But this is how shell works: a function definition is a compound command like any other and can occur pretty much anywhere. When it is encountered, it causes the function environment to be changed. For instance:

```shell
if ...; then
    ls () {
        echo hahaha
    }
fi
ls
```

is perfectly valid code. Depending on the execution of the condition, the last call to `ls` is either the function or the utility. Note that the converter wont be able to track down those uses. It will probably only allow for function definitions at toplevel. Still, one can intertwine function definitions with other pieces of code, like so:

```shell
my_function () {
    some code doing something
}

some code using 'my_function'

my_function () {
    some code doing something else
}
```

but I guess this could be rewritten (carefully, but it's not that hard) to something with all the definitions at the top of the file.

Anyway, it's probably cleaner to have all the function definitions at the beginning (that is, in the `program` type). I'll take care of the conversion.

Niols:
[...]

PR #34, opened 2018-10-19
Add functions, calls, shift, and argument expressions

benozol:
This PR suggest the declaration of functions in a semi-colon-separated list before the program instruction. @claudemarche do you have suggestions for a better syntax?
```
function f echo ['Hello'; arg 0 '!'];
function twice begin
  call f [arg 0];
  call f [arg 1]
end;
call twice ['colis'; 'world']
```

@Niols Could you please check `fromShell` – I am acting blindly in Morsmall. You can apply your changes to this branch.

When discussing the addition of functions I neglected the complexity of including the return-behaviour. Please have a look (esp. specification and implementation of `INot`

Targets #27


claudemarche:
I suggest no semi-colon, but a mandatory begin .. end for the main 
program, and also for the function bodies


function f
begin
   echo ['Hello'; arg 0 '!']
end

function twice
begin
   call f [arg 0]; call f [arg 1]
end

begin
  call twice ['colis'; 'world']
end

Le 19/10/2018 à 11:41, Benedikt Becker a écrit :

Niols:
The changes in `fromShell` look OK to me. I'll take care of the conversion of functions in an other PR later.

On a completely different topic: do you really want to call it `ICallBuiltin`? As long as it is absolutely clear that these correspond in Shell to **all the non-special builtins and other utilities that are not necessarily built in**, it's OK. I agree that, technically speaking, they are "built in the CoLiS language" (although, in the case of the `--realworld` option, they would not) but I still think it's misleading. By the way, the POSIX standard does not call them that, it calls them "utilities" and says they may or may not be built in the interpreter.

Niols:
(about `--realworld`, I've added #35)

Niols:
Also, this PR breaks CI because:

```
File "src/concrete/semantics.mlw", line 230, characters 19-31:
This function may not terminate, it cannot be used as pure
Makefile:19: recipe for target 'extract-why3' failed
```

benozol:
> I suggest no semi-colon, but a mandatory begin .. end for the main program, and also for the function bodies

@claudemarche I agree and will update the PR

> do you really want to call it ICallBuiltin?

Not really at all. What would you suggest is more suitable? `ICallUtility`?

> Also, this PR breaks CI because: This function may not terminate

The development version of Why3 correctly recognizes the termination criterion. We should probably pin Why3 by adding  `RUN opam pin why3 git@gitlab.inria.fr:why3/why3.git` to the Dockerfile?

src/concrete/semantics.mlw
```diff
@@ -319,17 +346,30 @@ module Semantics
...
+    (sta'', b) = interp_builtin sta' id ss ->
     let bhv = inject_behaviour inp b in
-    eval_instruction inp ctx sta (ICall s es) sta'' ctx bhv
+    eval_instruction inp ctx sta (ICallBuiltin id es) sta'' ctx bhv
+
+  | eval_call_function: forall inp ctx ctx' sta sta' sta'' id es ss ins bhv.
+    eval_list_expr inp ctx sta es sta' ss ->
+    ctx.fenv id = Some ins ->
+    eval_instruction { inp with arguments = ss } ctx sta' ins sta'' ctx' bhv ->
+    let bhv' = match bhv with BNormal b | BReturn b -> BNormal b | BExit b -> BExit b end in
```
benozol:
```suggestion
    let bhv' = match bhv with BReturn b -> BNormal b | _ -> bhv end in
```

Niols:
I'm fine with pinning Why3 in the Dockerfile if you'd rather do that.

benozol:
> I'm fine with pinning Why3 in the Dockerfile if you'd rather do that.

OK let’s do that in PR #26 when really necessary because the issue you mentioned was easily to circumvent

Niols:
> > do you really want to call it ICallBuiltin?
> 
> Not really at all. What would you suggest is more suitable? `ICallUtility`?

It's not really nice, but it would avoid the confusion imo.

benozol:
Alright, I renamed "builtins" to "utilities" by commit 412d2dcd250c3f3102694e45155b35ff44901bbe in branch `functions`.

Because I am working from a library that blocks port 22/SSH, I pushed to `ssh.github.com` (that’s [pretty useful!](https://help.github.com/articles/using-ssh-over-the-https-port/)). But this PR is not updated and CI is not triggered :/ So let’s wait a bit longer…

Anything else for this PR? Here is the full diff: https://github.com/colis-anr/colis-language/compare/functions

Niols:
Hey, I received your mail but I can't see your message on the interface… We can see the commit in the PR and I triggered a Travis build by hand to see if it's fine. I don't think there is anything to add, so I can merge that as soon as the CI comes green, would you be OK with that?

benozol:
Alright!

Niols:
The arguments number is a choice to make, I don't think there's a better alternative. The programmers would be happy to say that the first argument is 0, but the Shellers would be happy to keep it starting at 1. Just fix something and document it somewhere (at least in the notes?)

Niols:
One should also add the oracle/meta/whatever file to the test Claude has just added (when the question of the indices is closed.)

benozol:
> Just fix something and document it somewhere

Choosing zero-indexed arguments would require changing the Shell/CoLiS conversion (Shell `$n` → CoLiS `arg (n-1)`). So adding the callee as the first argument may be easier but I am not sure if `$0` contains the name of the interpreter or the name of the script file (outside functions). Do you know if this is well-specified?

> One should also add the oracle/meta/whatever file to the test Claude has just added

Claude only added the (one-indexed) shell version of the test `function` that already has a meta file, which breaks the CI

> I'll take care of the conversion of functions in an other PR later.

Well, now that Claude added the shell test, we would also need the conversion for this PR to pass the CI. @Niols do you think you could add it to this PR?

benozol:
@Niols, another question: If I understand the [POSIX specs](http://pubs.opengroup.org/onlinepubs/009604599/utilities/shift.html) well, the return value of `shift` not zero if the argument is larger than `$#`.  OK. But do you think there is reasonable explanation for dash’s behaviour? (exiting even in unstrict mode)

```shell
$ dash -c 'set +x; f() { shift 2; echo hello; }; f 1'
dash: 1: shift: can't shift that many
```

benozol:
I have updated the PR with the `shift` instruction, and arguments with shell-like semantics.

Currently only the import from shell function fails (I think) @Niols, would you suggest to disable `tests/function.sh` for now, or can you add the conversion of shell functions to the PR?

Niols:
I'll add the functions to the converter. I wont handle complex things like `$@` yet, but that should be enough for a merge and some tests.

Niols:
Hey @benozol, look at what I found in the standard (*2.9.1 Simple Commands*):

> If there is no command name, but the command contained a command substitution, the command shall complete with the exit status of the last command substitution performed. Otherwise, the command shall complete with a zero exit status.

I think it settles the question we were asking ourselves a while ago about behaviours of string and list expressions

benozol:
Yes, again a beautiful standard. I am not sure however if it specifies the previous return value within the substitutions of a word. Below is a test for this with different results for `Y/$?` in dash and bash. Do you have an idea how to translate it into Colis (-compatible Shell) for a test case?

```shell
aux() {

    # Test return value of empty simple command
    true
    $(false)
    echo X/$?

    # Test return value within substitutions
    true
    echo $(false)$(echo Y/$?)  # bash: Y/1, dash: Y/0
    echo Z/$?
}

aux && true
```

benozol:
Nice. I have just fixed the test which was garbage. But it still does not pass because the test uses `shift`, which is already implemented in the concrete interpreter. Do you think we should care about importing `shift` from Shell in this PR? Or should we split the test `function.{cls,sh,meta}` and have only the colis script for the part that uses `shift`?

benozol:
A just in time ☺

Niols:
OK, I think I got the functions conversion working. One more working hypothesis:
- one does not define a function with a name that has been used before in the script

One other way would probably be to rename the functions on the fly to ensure that the names are unique, but I'm not sure it's necessary.

Should we add the "return"? Or not for now?

Niols:
About `$?`, you're right, it does not really specify this case. In our case, the first test should not happen because we want non-empty simple commands. However, the second test might happen and I don't know about it.

benozol:
`return` is already defined in the colis language and interpreter – so it would be nice to add it in the conversion (and a test I guess).

benozol:
>  `$?`…  the second test might happen and I don't know about it.

For now, the interpreter goes after bash and `$? holds the result from any preceding instruction, even under substitution. This is somewhat arbitrary (marginally easier implementation) but can be changed if we find more evidence for the other behaviour.

Niols:
Oh, I didn't know about return. It is now supported in the translation

Niols:
But one should add a test, I'll do that

benozol:
Rebasing this on `master` is surprisingly difficult because of the premature code for symbolic execution that slipped in via #36/#37. I will create a commit that removes the code for symbolic execution from `master` and rebase this PR on that, OK?

PR #36, opened 2018-10-19
Proposition for the interface of constraints

Niols:
Hello,

Here is a proposition for an interface for constraints. This PR would merge in the start-symbolic branch and thus directly in #31.

The goal of the interface is to be rather close to the LaTeX document giving specifications of commands:
- the low level constraints working on one feature are hidden,
- higher level constraints working on couples var/path are provided in the interface. These constraints correspond quite closely to what's written in the document of the specifications
- the specifications are rewritten accordingly. I personally find them nice to read:

```ocaml

let interp_mkdir sta args =
  match args with
  | [full_path] ->
     let full_path = Path.from_string full_path in
     let path, feat = Path.split_last full_path in
     Specification.apply_to_state sta @@
       fun root root' ->
       [ (* The dir "path" exists but does not have "feat". *)
         { pre_state = dir (root, path) >=> abs (root, path) feat ;
           outcome = true ;
           post_state = sim1 root full_path root'
                        >=> dir (root', full_path) >=> empty (root', full_path) } ;
         (* The dir "path" does not exist. *)
         { pre_state = ndir (root, path) ; (*FIXME: see clause.mli*)
           outcome = false ;
           post_state = eq (root, Path.empty) (root', Path.empty) } ;
         (* The dir "path" exists and has "feat". *)
         { pre_state = dir (root, path) >=> nabs (root, path) feat ;
           outcome = false ;
           post_state = eq (root, Path.empty) (root', Path.empty) } ]
  | _ ->
    [error "mkdir: not exactly one argument" sta]
```

There are *at least* a few points to discuss, I'll put comments in the files. And I expect your review for such an important matter :-)


src/constraints/clause.mli
```diff
@@ -1,107 +1,83 @@
...
+(** [reg x\[p\]] means "the path [p] exists in [x] and what's there
+   isn't a directory." *)
 
-val neq : Variable.t -> Variable.t -> t -> disj
-(** [neq x y c] returns the disjunction corresponding to [x != y and
-   c]. *)
+(* FIXME: We need a notation for "either [t] doesn't exist or it is
+   not a directory". This is the subtle difference between (¬dir) t
+   and ¬(dir t). Maybe things like "ex_and_dir", "ex_and_ndir",
+   "nex_or_ndir"? *)
```
Niols: This remark is very important. Actually, a lot of names could be clearer and we can probably detach ourselves (a little, considering that the specifications will be written in this document) from the logic behind to find more suitable names.

src/constraints/clause.mli
```diff
@@ -1,107 +1,83 @@
...
-val sim : Variable.t -> Feature.Set.t -> Variable.t -> t -> disj
-(** [sim x fs y] returns the disjunction corresponding to [x ~fs y and
-   c]. *)
-
-val nsim : Variable.t -> Feature.Set.t -> Variable.t -> t -> disj
-(** [sim x fs y] returns the disjunction corresponding to [not (x ~fs
-   y) and c]. *)
+val sim1 : Var.t -> Path.t -> Var.t -> conj -> disj
+(** [sim1 x p y] means, with [p = q/f] "the path [q] exists in both
+   [x] and [y] and [x] and [y] may only differ in the name [p]." *)
```
Niols: One probably still need `val sim : Var.t -> Path.t list -> Var.t -> conj -> disj` as, for instance, `sim x [p; q] y` is different from `sim1 x p y >=> sim1 x q y`.

src/constraints/dune
```diff
@@ -0,0 +1,6 @@
+(include_subdirs no)
+
+(library
+ (name constraints)
+ (public_name colis-language.constraints)
```
Niols: I separated the constraints from the main library, in order to make it more modular. For the rest of the project, it behaves as any lib that would happen to be called `colis-language.constraints`: you have to include it as a dependency in `src/dune` and then you get access to the main module only (that is `Constraints`). Also, there should be a specific entry point in the documentation for this sub-library.

src/symbolic/symbolicBuiltin.ml
```diff
@@ -81,102 +24,103 @@ let with_clause sta clause =
...
-  | f :: p -> String.join "/" (List.rev p), f
-  | _ -> failwith "last_path_component"
+module Specification =
+  struct
+    open Clause
+
+    type case =
+      { outcome : bool ;
+        pre_state : conj -> disj ;
+        post_state : conj -> disj }
```
Niols: I've not included what's printed and where, maybe these could only be strings?

Niols:
Also, the CI is broken because it is in #31. And that's because of the "batteries" lib that #31 now includes in the project. I think the way to fix this would be to:
- merge `master` into `start_symbolic`
- update `colis-language.opam` in `start_symbolic` to add batteries as a dependency
- check that the CI passes in `start_symbolic` (and thus in #31), fix things if needed
- merge `start_symbolic` into `start_symbolic_constraints`
- check that the CI passes in `start_symbolic_constraints` and thus in this PR, fix things if needed

Or we can just merge everything in master and see afterwards ;-)

Niols:
@benozol I'll copy things from your e-mail here and answer to them, is that OK for you?

src/constraints/clause.mli
```diff
@@ -1,107 +1,83 @@
...
-(** The type of disjunctions of clauses. *)
+val exists : (Var.t -> conj -> disj) -> conj -> disj
 
-val top : disj
-(** The disjunction containing only the empty clause (that is,
-   "true"). *)
+val fold : ('a -> conj -> 'a) -> 'a -> disj -> 'a
+val map_to_list : (conj -> 'a) -> disj -> 'a list
+(* FIXME: this is exactly (when you reverse the arguments) the monadic
+   bind. But I don't dare writing it that way. *)
```
Niols: > Why not ;) ? Or why not just `val disjuncts : disj -> conj list`? (Is this a proper name of the operands of a disjunction?) And additionally indicating that all resulting conjunctions are satisfiable.

Because it gives a glimpse of the structure that's behind. And if we change the structure behind (from lists to sets for instance), this would stop being the monadic bind. I guess the best solution is indeed to add a function to expose the conjunctions, like you suggest.

src/constraints/clause.mli
```diff
@@ -1,107 +1,83 @@
...
 
-val bottom : disj
-(** The empty disjunction (that is, "false"). *)
+(** {2 Monadic functions} *)
 
-val sat : disj -> bool
-(** [sat d] checks whether [d] is a satisfiable disjunction. This is
-   infact only testing whether the disjunction is non-empty. *)
+val pure : conj -> disj
```
Niols: > The name `pure` suggests absence of side effects but I guess that’s not what is meant here. `return` or `unit` may be more neutral… how would you call it without thinking monadically? `singleton`? `singleton_conjunction`? `disj`?

Indeed, pure is not a good name here. I guess there are two ways to see it:
- it's a function that takes `conj` and produces the `disj` containing only this `conj`. In which case, `singleton`, `disj`, etc. would make sense
- its a "constant" of type `conj -> disj`. In which case, it is in fact true…

I don't know which one to choose. Maybe the two names? Be cause it makes no sense to apply "true" to something, so `singleton` might be better there. However, when writing a specification of type `conj -> disj` for `echo`, for instance, one wants to write just `true` and not `singleton` that would mean nothing to the reader…

src/constraints/clause.mli
```diff
@@ -1,107 +1,83 @@
...
+val (>=>) : (conj -> disj) -> (conj -> disj) -> (conj -> disj)
 
-val exists_comp : Variable.t list -> t -> t
-(** [exists xs c] returns the clause corresponding to [exists Cxs. c]
-   where [Cxs] represents the complement of [xs]. *)
+(** {2 High-level functions working on terms} *)
 
-val exists_comp_d : Variable.t list -> disj -> disj
-(** Same as {!exists_comp} but on disjunctions. *)
+type term = Var.t * Path.t
```
Niols: > A `Path.t` may contain components `.`, and `..`. Will those functions that take a `term` or `Path.t` always take care of such path components and will they deal with symbolic links during resolution?

You are right, I was going a bit fast there. We want two kinds of pathes: resolved (or whatever their name is) and "usual" ones that may contain `.`, `..` (and even symbolic links, but I thought we had abandoned hope for these?). And in fact, a term should be:
- a variable for the root
- a path for the current working directory
- the path where we work

The definition of a function working with a term (appart for negated versions) is always that the term resolves (and there, we need to take care of these path components) to a variable where the lower function of the same name applies. So the path component would be taken care of there.

I wonder how we want to handle the CWD though. Is it only a variable in the interpreter that concatenates it to the path before giving the path to the constraint? Or is it handled in the terms? (then a term would take one variable, one resolved path and one path).

src/constraints/clause.mli
Niols: > Will the negative version of the functions below correspond to the error cases of the symbolic path resolution in section 8.2.3 of the document "Specification of UNIX Commands"?

Yes, just like in this document. In the document, we have basic constraints, like `reg` and `nreg` that take a variable and specify that it must be a regular file (or not be a regular file). Then we lift them to terms (that is a pair variable/path) and the semantics of the lifted `reg` is "the path exists, resolves to a variable and this variable must be a regular file" (that is, `reg(r[p]) = \exists x. resolves(r, p, x) and reg(x)`).

But then, although `nreg(x) = ¬ reg(x)`, we cannot say the same for their lifted versions, as:
- `nreg(r[p])` means that "`p` resolves in `r`, gives a variable that is not a regular file",
- `¬ reg(r[p])` means that NOT "`p` resolves in `r`, gives a variable that is a regular file".

And we want to use them both! For instance:
- the success case of `test -f p` is `reg(r[p])`, its failure case `¬ reg(r[p])`,
- one success case of `mkdir -p p` is `dir(r[p])`, its failure case `ndir(r[p])`

src/constraints/clause.mli
```diff
@@ -1,107 +1,83 @@
...
+val nfen : term -> Feat.Set.t -> conj -> disj
+(** [fen x\[p\] fs] means "the path [p] exists in [x] and what's there
+   does have a feature that is not in [fs]." *)
 
-val abs : Variable.t -> Feature.t -> t -> disj
-(** [abs x f c] returns the disjunction corresponding to [x\[f\]^ and
-   c]. *)
+val empty : term -> conj -> disj
+(** [empty x\[p\]] means "the path [p] exists in [x] and what's there
+   does not have any feature." *)
```
Niols: > = `fen x Feat.Set.empty y`?

yes

src/constraints/clause.mli
Niols: > Agreed. But out of interest: Is this required for implementing the builtins/utiltities?

It is for utilities that work at two different pathes of the file system, like `mv` and `cp`.

src/constraints/dune
Niols: > Fine. What about keeping libraries that we envisage to separate from the main project at one point (`constraints`, `morsmall`) in a directory `lib/` instead of `src/`?

That's fine with me and is only a `mv` considering how Dune works. Removing them will be also rather simple, we'll just have to replace the dependency in `colis-language.constraints` by the new name.

src/symbolic/symbolicBuiltin.ml
Niols: > That’s fine, these may serve at once as documentation of the cases and at some point for tracing the symbolic execution.

What do you mean? I was thinking about what may be printed by the utility that could go in a specification case.

src/symbolic/symbolicBuiltin.ml
```diff
@@ -81,102 +24,103 @@ let with_clause sta clause =
...
+           outcome = true ;
+           post_state = sim1 root full_path root' >=> reg (root', full_path) } ;
+         (* The dir "path" exists and has "feat". *)
+         { pre_state = dir (root, path) >=> nabs (root, path) feat ;
+           outcome = true ;
+           post_state = eq (root, Path.empty) (root', Path.empty) } ;
+         (* The dir "path" does not exist. *)
+         { pre_state = ndir (root, path) ; (*FIXME: see clause.mli*)
+           outcome = false ;
+           post_state = eq (root, Path.empty) (root', Path.empty) } ]
```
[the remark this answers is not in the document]
Niols: Also, I was thinking: we will probably always write `Specification.apply_to_state sta @@ fun root root' ->`. So maybe, the functions `interp_touch` should only be:

```ocaml
let interp_touch sta args = Specification.apply_to_state sta (spec_touch args)
```

with:
```ocaml
let spec_touch args root root' =
  <the whole list of specifications here>
```

And then, we could probably separate the place where we write **only** specifications that look like the paper and the place where we have all this machinery. That would allow other to review only the specifications without wondering what all these higher-order functions are ^^

src/symbolic/symbolicBuiltin.ml
```diff
@@ -81,102 +24,103 @@ let with_clause sta clause =
...
-    in
-    let fails =
-      noresolve sta.filesystem path >>=
-      with_clause {sta with result = true}
-    in
-    oks @ fails
-  | _ -> [error "test-e: Not exactly one argument" sta]
+(* Interpretation of all commands *)
 
 let interp (sta: state) (cmd: string) (args:string list) : state list =
```
Niols: Infact, `interp` could maybe be:

```ocaml
let interp (sta: state) (cmd: string) (args: string list) : state list =
  match Utilities.find_specification_opt cmd with
  | None -> [error ("Unknown builtin: " ^ cmd) sta]
  | Some spec -> Specification.apply_to_state sta spec
```

if all the specifications are given using the same format in an other file?

Niols:
Hey @benozol. I've written a lot of sutff. I'm ready to give it an other iteration, but there are a few things I'd like to have your opinion on first. Also, an other idea that I had: have a module `Clause` containing all the constraint solving and stuff and providing one guarantee that if one manipulates `Clause.conj`, then this is a satisfiable conjunction. And then, the module to be used (`Constraint`?), with a signature like:

```ocaml
type t

val feat : (Var.t * Path.t) -> Feat.t -> t
val abs : (Var.t * Path.t) -> Feat.t -> t
...
val fen : (Var.t * Path.t) -> Feat.Set.t -> t
...

val (&) : t -> t -> t
(** Conjunction of constraints. This is in fact the list monadic composition but shhhhh *)

val add_to_clause : t -> Clause.conj -> Clause.disj
(** This is in fact the polymorphic identity but shhhh *)
```

So that readers have a more usual interface for a language that allows to write constraints. This interface almost looks lower-lever than the one in `Clause`. It looks like these are just constructors for a simple variant type. I personally like it. What do you think?

src/constraints/clause.mli
benozol: Ah right, `conj -> disj` is `true` ! But to my understanding this interpretation makes (only) sense in the context of monadic composition `(true : conj -> disj) >=> …`, where it `true` is also redundant. By contrast it *is* used to inject a single clause into the monad mound by `>>=`:

https://github.com/colis-anr/colis-language/blob/36f400b18164eaffab7e2be24e92781f04089c2e/src/symbolic/symbolicBuiltin.ml#L36-L40

So I guess the name should rather be chosen to describe its use in the latter usecase…

src/constraints/clause.mli
benozol: > symbolic links, but I thought we had abandoned hope for these?

Oh I thought they were still a thing – maybe we should discuss with the others?

> The definition of a function working with a term is always that the term resolves

OK that’s good (and we discuss the negated constraints in the conversion starting with your `FIXME`…)

> 3. the path where we work / I wonder how we want to handle the CWD though.

What about requiring terms in `Clause` to contain an absolute path, and a function `Path.split : cwd:Feature.t list -> s:string -> t` that prepends `cwd` to `s` if `s` is relative?

src/constraints/clause.mli
benozol: OK then I agree even more with suggestion to choose names for the lifted constraints that reflect this added behaviour ("ex_and_dir", "ex_and_ndir", "nex_or_ndir")

src/constraints/clause.mli
benozol: Very well.

src/constraints/dune
benozol: OK then we know that everyone can do changes in `src/` and only the responsible of the the other libraries (uhm, @Niols ) in `src/`.

src/symbolic/symbolicBuiltin.ml
benozol: It should not be printed (at least not on `state.stdout`) because that wouldn’t correspond to the specification. To collect these doc strings for cases during symbolic execution may help, however, to figuring out how we ended up in a given state.

src/symbolic/symbolicBuiltin.ml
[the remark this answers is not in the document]
benozol: > `which` that tests an existence and prints something

> we could probably separate the place where we write only specifications that look like the paper and the place where we have all this machinery

Perfect, that’s what I meant. Ideally we even separate the mapping from the arguments of `inter`/`interp_touch` (of type `string list`) and `spec_touch`. But we should also keep in mind that different arguments can call very different specifications (`test -e p` vs `test -n p` or `rm p` vs `rm -rf p`)

src/symbolic/symbolicBuiltin.ml
benozol: Ah then it would really make sense/require adding the utilities stdout to the `Clause.Specification.case`

src/constraints/clause.mli
Niols: It might also be used for command specification cases when they do nothing. In this case, we need to represent the constraints that does nothing, and with need the `true`. So why not both the names?

src/constraints/clause.mli
Niols: > > symbolic links, but I thought we had abandoned hope for these?
> 
> Oh I thought they were still a thing – maybe we should discuss with the others?

Things is, I think we just can't handle them. Because then, when resolving a path in the filesystem that is not known, each path component can in fact be a symbolic link to anywhere (including non-existing parts of the filesystem)… But yes, we should discuss with the others.

> > 1. the path where we work / I wonder how we want to handle the CWD though.
> 
> What about requiring terms in `Clause` to contain an absolute path, and a function `Path.split : cwd:Feature.t list -> s:string -> t` that prepends `cwd` to `s` if `s` is relative?

In the symbolic engine (or the specifications) then? However, it can be an absolute path but not a resolved one. The constraint for `reg[a/../b]` is not the same as `reg[b]`.

Niols: But that means that constraints only care about absolute pathes and ignore the cwd. That's probably nice as it does not have much to do with them.

benozol: > when resolving a path in the filesystem that is not known, each path component can in fact be a symbolic link to anywhere 

AFAIK @claudemarche’s compromise was to treat a file as symbolic link only when this is *known* in the filesystem.

> But that means that constraints only care about absolute pathes and ignore the cwd. That's probably nice as it does not have much to do with them.

Exactly, otherwise you would first have to resolve the CWD and the relative path, which should have the same result, right?

benozol:
I like the proposed interface of `Constraint`. And on that level it makes sense to give another name to `true` because the underlying type is abstract (or is this already `add_to_clause`?).

> guarantee that if one manipulates Clause.conj, then this is a satisfiable conjunction

What do you mean?

src/constraints/clause.mli
claudemarche: > > when resolving a path in the filesystem that is not known, each path component can in fact be a symbolic link to anywhere
> 
> AFAIK @claudemarche’s compromise was to treat a file as symbolic link only when this is _known_ in the filesystem.

I confirm. First I don't see another solution: how would you express a constraint expressing that a node would be a link to *any* path? Second: it is a good compromise. Drawback: in that case we are doing an under-approximation of executions, so we lose what I called before the "soundness"

> > But that means that constraints only care about absolute pathes and ignore the cwd. That's probably nice as it does not have much to do with them.
> 
> Exactly, otherwise you would first have to resolve the CWD and the relative path, which should have the same result, right?

Not sure I get the source of the discussion. But for me, for sure, constraints are not concerned about CWD. Symbolic path resolution is concerned by CWD, though.

Niols:
Hum, we are in a module clause that defines `Clause.conj` and `Clause.disj`. You'd add `Clause.clause`? Or `Clause.t`? I'd like to have in the types the difference between:
- `conj`: satisfiable conjunctions
- `disj`: disjunction of satisfiable conjunctions
- `raw_conj`: conjunction of literals, not necessarily sat (because the computation hasn't taken place yet)

Actually, `raw_conj` is not exactly a conjunction as we can write the disjunction with this type:

```ocaml
let (++) r1 r2 = fun c ->
  (r1 c) @ (r2 c)
```

so it's indeed a "clause" that is not necessarily sat. I was thinking about something like `sat_conj`, `sat_disj` and something else, but I think that `sat_disj` is misleading. Maybe with good documentation, we could have `Clause.t` representing any clause, and then `conj` and `disj` representing only satisfiable conjunctions and disjunctions of satisfiable conjunctions? I'd like that in term of interface, but I still need names for these conjunctions and disjunctions…

Niols:
So, now, the tool builds (I've added a dummy implementation made only of `assert false`) but the tests fail because of "test-symbolic" that does not have a `.meta` file.

```
$ make test
dune runtest
   run_tests alias tests/runtest (exit 2)
(cd _build/default/tests && runner/run_tests.exe)
Fatal error: exception (Sys_error "./test-symbolic.meta: No such file or directory")
make: *** [test] Error 1
Makefile:8: recipe for target 'test' failed
The command '/bin/sh -c eval $(opam env) && make test' returned a non-zero code: 2
```

Where does it come from? Should we add a `.meta` file for this test or is there something else that we expect from it?

benozol:
> Where does it come from? Should we add a .meta file for this test or is there something else that we expect from it?

It comes from branch `start-symbolic` where it servers as a reminder that symbolic execution should be tested (as long as we haven’t proven it…). We could also do the tests in another PR and delete it in `start-symoblic` for now?

Niols:
I don't really like having the tests failing because "we'll fix them someday". It breaks the principle of the CI that is that as soon as you break something, you get informed. I'd put this "we should test symbolic execution" either in an other PR or in an issue

benozol:
Well `start-symbolic` is really WIP and I didn’t expect anything to pass. I have deleted the test file, and the failing CI file should be fixed by this PR then, right?

Niols:
Yes, it should be. Do you think there is more we need to add to this PR? Or should we merge it into start-symbolic?

benozol:
I think the PR is fine and can be merged

Niols:
I would just like to replace the implementations of `resolve`, `noresolve` and `sim1` to support any path. That shouldn't be too long, but I'll need your review on that.

Niols:
@benozol So do you think we should merge that into #31? Or into master? This PR targets directly the *start-symbolic* branch, so that would all appear in #31, but I can change its target.

lib/constraints/interface.ml
```diff
@@ -0,0 +1,227 @@
...
+
+  val and_ : t -> t -> t
+  (** The conjunction of two [t]. *)
+
+  val (&) : t -> t -> t
+  (** The conjunction of two [t] (alias for {!and}). *)
+
+  val or_ : t -> t -> t
+  (** The disjunction of two [t]. Beware! [(++)] has a higher
+     precedence than [(&)]. *)
```
benozol: The second half of the comment can be removed

Niols:
This PR still works by resolving paths hidden in the constraint interface indeed. I wanted to wait for the PDF document to be stable before changing the interface. The impact of one decision or the other should not change much to the implementation so I think we can just postpone that until the discussion is really closed (I see Ralf tomorrow), what do you think?

benozol:
Yes that’s fine.

Niols:
So do I rebase this into master?

benozol:
Yes. if the changes resulting from further discussion with Ralf only require replacing some lines, this could be done in subsequent commits. And #31 would something to build upon.

Niols:
Alright, let's do this.

PR #40, opened 2018-11-07
Restart symbolic execution

benozol:


benozol:
OK at least symbolic execution of shell scripts is doing something now, using the new constraint interface.

The PR is currently targeting branch `skdjfh`. This should be changed to `master` once the former is merged.

Niols:
An other way would be to merge this into #41 and then to merge #41 into master. I don't know what would be the best scenario

benozol:
@Niols symbolic execution is starting to be fun:

```
$ cat test.sh
mkdir abc
$ ./bin/colis --run-symbolic test.sh
* Initial state
- root: root₁
  cwd: /
  clause: dir(root₁)
* Success states
- root: root₁₂
  cwd: /
  clause: ∃ abc₅⋅ root₁₂[abc]abc₅ ∧ root₁[abc]↑ ∧ dir(root₁) ∧ dir(root₁₂) ∧ dir(abc₅) ∧ abc₅[∅] ∧ root₁ ~{abc} root₁₂
* Failure states
- root: root₁₂
  cwd: /
  clause: root₁ = root₁₂ ∧ dir(root₁) ∧ ¬ dir(root₁)
- root: root₁₂
  cwd: /
  clause: ∃ ?₇⋅ root₁ = root₁₂ ∧ root₁[abc]?₇ ∧ dir(root₁)
```
However, I would have expected that the contradiction in the second failure state was detected by the naive solver. Not yet?

Niols:
In the first failure case, you mean? I know where it comes from: I have implemented only the IJCAR paper for now, which means that `dir` and `reg` are just stupid for now. I'll get them working before the meeting

Niols:
Why is the CI failing? It seems to fail replaying the proof. Is this because of what you told us earlier today?

benozol:
There seem to be two separate issues for attaching the proofs while replaying: minor changes in the preconditions to make them compatible with symbolic execution, and using the new dev’ version of Why3 for replaying the old proofs (this shouldn’t be a problem for the CI that uses still the same version).

I will have another look next week.

Niols:
Oh, alright. so the problem, for the CI's point of view, would more be the changes in the preconditions, right? Since you pinned Why3 dev's commit

benozol:
We should also create a test suite for symbolic execution.

But the full specification of result states including formulae describing the filesystem seems to be a lot of effort and also quiet fragile with respect to variable changes in the formulae.

I see two alternative approaches:

1. Introduce a notion for concrete filesystems that can be tested for correspondence with an feature tree formulae. The success case of the test script `mkdir x/y` could then something like `before: /x(dir)` and `after: /x/y(dir,[∅])`.
2. Currently, each case of a utility prints a debug message that describes its action (or reason for non-action). It may be sufficient to specify test results for symbolic execution as a set of pairs of result (success/error) with the output (including the debug messages from the utilities)?

Any thoughts? @Niols

benozol:
Sure, we are aiming at proving some properties of the symbolic execution engine itself (over- and underapproximation). But these proofs won’t state anything about the behaviour of the utilities because they do not have a concrete semantics. So testing the implementation of utilities seems necessary but maybe isolated in terms of our recent discussion in #41).

Niols:
Can't we just test the specification against the actual utility then? To test all the utilities in an independant way (in addition to testing the whole symbolic engine).

(God, if we implement all these tests, we'll need heavy documentation about them, even for us)

Niols:
Hey, two questions:

1. Do you quantify on intermediary roots now? If yes, have you tried the ocaml-nox script again?

2. Is there a way in the current implementation of the specifications to specify a command as a small CoLiS script? It's kind of a circular dependency, but it will be required as some commands are specified like that.

benozol:
>     1. Do you quantify on intermediary roots now? If yes, have you tried the ocaml-nox script again?

Yes indeed! It finds in less than a  that there are no runtime errors in the ocaml-nox [1] script but 1,225 ways to succeed☺ assuming only a few standard directories (we should actually assume the FHS when running scripts)

>     2. Is there a way in the current implementation of the specifications to specify a command as a small CoLiS script? It's kind of a circular dependency, but it will be required as some commands are specified like that.

Currently not. This means tight interaction between utilities and functions, which we have to think through. Also, the CoLiS language is not yet expressive enough for the scripts from the specification, right?

[1] 
```shell
if ! test -e /usr/local/lib/ocaml; then
    if mkdir /usr/local/lib/ocaml 2>/dev/null; then
	      echo chown root:staff /usr/local/lib/ocaml
	      echo chmod 2775 /usr/local/lib/ocaml
    fi
fi
if ! test -e /usr/local/lib/ocaml/4.01.0; then
    if mkdir /usr/local/lib/ocaml/4.01.0 2>/dev/null; then
	      echo chown root:staff /usr/local/lib/ocaml/4.01.0
	      echo chmod 2775 /usr/local/lib/ocaml/4.01.0
    fi
fi
if ! test -e /usr/local/lib/ocaml/4.01.0/stublibs; then
    if mkdir /usr/local/lib/ocaml/4.01.0/stublibs 2>/dev/null; then
	      echo chown root:staff /usr/local/lib/ocaml/4.01.0/stublibs
	      echo chmod 2775 /usr/local/lib/ocaml/4.01.0/stublibs
    fi
fi

for i in /usr/lib/ocaml/3.06 /etc/ocaml /var/lib/ocaml; 		\
    do									\
	      if test -e $i/ld.conf; then					\
		        echo "Removing leftover $i/ld.conf";			\
		        echo rm -f $i/ld.conf;					\
		        echo rmdir --ignore-fail-on-non-empty $i;			\
	      fi;								\
    done
```

src/symbolic/utilitiesSpecification.ml
```diff
@@ -77,6 +77,6 @@ type specifications = cwd:Path.t -> root:Var.t -> root':Var.t -> case list
 
 let under_specifications : specifications -> state -> (state * bool) list =
   fun spec state ->
-    let new_root = Var.fresh ~hint:(Var.to_string state.filesystem.root) () in
+    let new_root = Var.fresh ~hint:(Var.hint state.filesystem.root) () in
```
benozol: @Niols don’t worry any more about the variable indexes  suitable for counting atoms in the universe :)

Niols: Ooh, that's why! Do they look alright now?

Niols: I mean, what's the counter for the ocaml-nox for instance?

Niols:
> > 1. Do you quantify on intermediary roots now? If yes, have you tried the ocaml-nox script again?
> 
> Yes indeed! It finds in less than a that there are no runtime errors in the ocaml-nox [1] script but 1,225 ways to succeed☺ assuming only a few standard directories (we should actually assume the FHS when running scripts)

So much suspense! Less that a what? ;-)
Also, 1225 execution pathes is way too much. Any ideas about where it comes from?
 
> > 2. Is there a way in the current implementation of the specifications to specify a command as a small CoLiS script? It's kind of a circular dependency, but it will be required as some commands are specified like that.
> 
> Currently not. This means tight interaction between utilities and functions, which we have to think through. Also, the CoLiS language is not yet expressive enough for the scripts from the specification, right?

The CoLiS language is expressive enough to have small scripts generated by OCaml code used as specifications. For instance, although `mkdir -p` is not writeable in CoLiS, `mkdir -p a/b/c` is. By the way, it's probably something like

```shell
[ -d a ] || mkdir a
[ -d a/b ] || mkdir a/b
[ -d a/b/c ] || mkdir a/b/c
```

src/symbolic/utilitiesSpecification.ml
benozol: The numbers seem more realistic now: `r₂₀₆₈₅` for ocaml-nox. @claudemarche  suggested, however, to choose the indexes for printing *while printing* to keep them as low as possible.

benozol:
Less than… a minute!

I assumed that the large number of success states comes from the combinations of cases of different applications of `noresolve`. Plausible?

> The CoLiS language is expressive enough to have small scripts generated by OCaml code used as specifications

That may be feasible. We could change the OCaml function for interpreting the utiltiies that is exposed to the Why3 interpreter to return either

 - a set of states as it currently does  for “atomic utilities” (such as `mkdir` with a single feature argument), or
 - an instruction for evaluated like a dynamic function body for “complex utilities” (such as `mkdir -p` or `mkdir a b c`). The instruction can potentially refer to other atomic or complex utilities.

src/symbolic/filesystemSpec.ml
```diff
@@ -0,0 +1,86 @@
...
+      node "lib" [];
+    ];
+  ];
+  node "etc" [];
+  node "var" [
+    node "lib" [];
+  ]
+]
+
+(* TODO to be extended to FSH 3.0? *)
```
benozol: The execution time of `touch x` is more than 350 times longer when using `fsh` than when using `empty`. We may need a more efficient constraint solver to deal with large filesystems…

src/symbolic/utilitiesSpecification.ml
Niols: Yes, I had this idea too, but I didn't implemented it yet

src/symbolic/filesystemSpec.ml
Niols: We definitely need a more efficient constraint solver. However, I wont take care of this before the meeting. I'd like to debug everything on this naive system so that we can use it to test the efficient one.

Niols:
> Less than… a minute!

Argh, I was hoping for the second. We'll need something more efficient than the naive solver.

> I assumed that the large number of success states comes from the combinations of cases of different applications of `noresolve`. Plausible?

Plausible indeed, but I should look closely at the generated constraints to understand.

> > The CoLiS language is expressive enough to have small scripts generated by OCaml code used as specifications
> 
> That may be feasible. We could change the OCaml function for interpreting the utiltiies that is exposed to the Why3 interpreter to return either
> 
> - a set of states as it currently does  for “atomic utilities” (such as `mkdir` with a single feature argument), or
> 
> - an instruction for evaluated like a dynamic function body for “complex utilities” (such as `mkdir -p` or `mkdir a b c`). The instruction can potentially refer to other atomic or complex utilities.

It looks fine indeed. I guess that one would return a CoLiS AST. To be executed in a new environment (as if `/bin/mkdir_p` were a Shell script) or in the current one?

Niols:
I've been thinking about the numerous cases that the tool reports in ocaml-base-nox's postinst script. I counted them myself (well, not one by one) and I end up with 3936 cases. That is exactly the same number as the tool does. It's good news for the state of the tool but probably a bit worrying for the complexity.

Although in that case it comes from the fact that we run everything with no hypothesis. If we assume the existence of:
```
/usr/local/lib
/usr/lib
/etc
/var/lib
```
I think we have "only" 144 cases. Have you tried this (it might take forever…)?

benozol:
Almost !

Using `Filesystem_spec.empty` that contains no hypothesis about the FS, the symbolic execution end up with exactly 3936 cases after 55sec. Using `Filesystem_spec.simple` that encodes exactly these four directories, we end up with 576 cases after 37sec. (There isn’t yet an easy way to change the hypotheses but I’ll add it before tomorrow for the demo.)

Niols:
OK, I don't know how I counted, but I checked my notes and I indeed end up with 576. Dammit, it's really a lot… we're paying here the cost of the DNF

benozol:
Green lights! Any final words?

(This PR includes #41 and targets master.)

benozol:
@Niols ping

Niols:
How could I miss this message? You have my green light too! Let's get rid of these PRs.

PR #41, opened 2018-11-26
Naive implementation of constraints

Niols:
Hey @benozol, constraints are coming!

Sorry about the name of the branch. I pushed onto master, panicked, force pushed to fix it and saved my stuff to a random keystroke on my keyboard… You don't have to read all the code, of course. Here is a summary of what I changed/added:

### The basic idea of this PR

- The naive representation of a conjunction is just a set of literals (an atom with a positive/negative sign). A disjunction is a list of conjunctions.
- We add literals to the sets and call a function `normalize`. This function basically applies rules until we reach a normal form (the normal form can be the empty disjunction).
- A rule is a function of type `conj -> disj option` returning. It looks for a pattern in the conjunction. If the pattern is found, it makes changes and applies them to the conjunction, returning (possibly) a disjunction.

### What is missing, what should be improved:

- There are a lot of rules missing, I have to add them.
- There are only two tests, I should add many more.
- The pattern engine is probably highly inefficient, I could improve that (although, since the whole system of updating a pool of literals is inefficient, I'm not sure it's worth my time…)

### Other changes

- I moved the implementation of constraints from `lib/constraints` to `lib/constraints/src`. It will make future separation work easier and it allows to have a directory `lib/constraints/tests` containing constraints-specific tests. These tests are launched by Dune correctly and thus are part of the `make test` call.
- I used some deriving PPXes. It comes particularly handy when one wants monomorphic equality (that was my case) or debug printing of OCaml values.
- I included a logging utility, [Logs](http://erratique.ch/software/logs/doc/Logs.html#basics). It is pretty easy to use and I think that we should use it pretty much everywhere in the project to keep track of what is happening. A good practice is to have several parts of the project using distinct loggers. The constraints are using their own. Then, in the executable file, we can choose which loggers is set to which log level.
- By moving all that, I most certainly broke stuff in #40. I know for instance that you wanted to print formulas which is not possible in this PR yet. Since I didn't know the status of #40, I did not take care of that, but I can if you tell me what you need.


lib/constraints/src/implementation/affect.ml
```diff
@@ -0,0 +1,47 @@
+open Constraints_common
+
+type t = (* invariant: always sorted *)
```
yurug: Why do you need that?

lib/constraints/src/implementation/affect.ml
```diff
@@ -0,0 +1,47 @@
+open Constraints_common
```
yurug: This filename shoud `assignment.ml`, right?

lib/constraints/src/implementation/naive.ml
```diff
@@ -0,0 +1,253 @@
...
+     if equal_conj conj' conj then
+       None
+     else
+       (
+         Log.debug (fun m -> m "Rule %s applied" name);
+         Some [conj']
+       )
+  | Some disj' ->
+     Log.debug (fun m -> m "Rule %s applied" name);
+     assert (List.for_all ((<>) conj) disj');
```
yurug: I am not sure to understand why this assertion holds.

lib/constraints/src/implementation/affect.ml
Niols: Indeed

lib/constraints/src/implementation/affect.ml
Niols: Because I want a mapping from meta variables to other stuff. So I have to check when extending the mapping that it does not violate the previous one. I just replaced the list by `Metavar.Map`

lib/constraints/src/implementation/naive.ml
Niols: This one is a bit tricky, tbh. The IJCAR paper says the following:

> We say that such a rule *left => right* applies to a clause *c* if:
> 1. The pattern *left* matches the complete clause *c* modulo associativity and commutativity of conjunction.

This is the job of the `Pattern` module.

> 2. The side conditions of the rule, if any, are met.

This is the optional `pred` argument to `Pattern.find`

> 3. The transformation yields a formula which is different from *c*.

This is where it becomes annoying. The easy case is when the rule transforms a conjunction into an other conjunction. In that case, we just check the equality of the conjunctions (module associativity and commutativity, but since the literals are in a set, it's fine).

I don't really know what the paper means in case the rule yields a disjunction. We most likely want that none of the output conjunctions is equal to the input one, because that would create an infinite loop. However, when it is the case, should we consider that that is an error or that the rule simply does not apply?

"Luckily", there is no rule that may produce this (or I really missed something), so I went for the "safe" assertion (I'd rather have my program stop here than continue with a wrong constraints solver).

I don't know if there's something else to do here (at least, I don't have any idea for now). I should at least put a comment here.

Niols:
Hum, that was not my plan, but I can certainly do that. What do you prefer?

lib/constraints/src/common/path.mli
```diff
@@ -0,0 +1,22 @@
...
+
+val empty_rel : rel
+
+val split_first_rel : rel -> (comp * rel) option
+
+type t =
+  | Abs of rel
+  | Rel of rel
+
+(* val from_string : string -> t *)
```
benozol: Is there another function in the interface to "parse" a string to a `Path.t`?

lib/constraints/src/common/path.mli
```diff
@@ -0,0 +1,22 @@
+type comp =
+  | Up
+  | Here
+  | Down of Feat.t
+
+type rel = comp list
+
+val empty_rel : rel
+
+val split_first_rel : rel -> (comp * rel) option
```
benozol: Is the last path component of a string path `"a/b/c"` *on top* of the `rel` (`["c";"b";"a"]`)? Otherwise we need another function `first_last_rel`, e.g., for implementing `touch`

lib/constraints/src/common/var.mli
```diff
@@ -0,0 +1,17 @@
+type t
+
+val fresh : ?hint:string -> unit -> t
+
+val compare : t -> t -> int
+val equal : t -> t -> bool
+
+val show : t -> string
```
benozol: Better the same name for the string converter in `Feat` and `Var`? (is currently `Feat.to_string`)

benozol:
> Hum, that was not my plan, but I can certainly do that. What do you prefer?

Well, I have just started doing so in #40 but encountered a problems with the interface (see my other comment about `Path`). If you are ok with potentially more such requests, I’ll continue the update of `symbolicUtilities`.

Niols:
I really don't like the current `Path` interface, but let's get it working. I'm currently implementing the missing functions (`split_last_rel`, `from_string`, etc.)

Niols:
Here they are

benozol:
Thanks!

benozol:
I guess we also need `noresolve`

Niols:
Good point, and we don't want `¬ resolve`. On it!

lib/constraints/src/common/path.mli
```diff
@@ -0,0 +1,23 @@
...
+val split_last_rel : rel -> (rel * comp) option
```
benozol: For implementing the utilities it would be easier if those functions operated on `Path.t` instead of `Path.rel`

benozol:
This looks fine so far from the point-of-view of the utilities.

Do you plan to include a pretty-printer for `sat_conj`? This would be useful to for printing the result states in the `colis` command

lib/constraints/src/common/path.mli
benozol: Not sure either if there will be use of `split_first` for the utilities…

Niols:
Here it is…

lib/constraints/src/common/path.mli
Niols: OK, so I'll not put it there as I'm not really convinced about how it should behave on all pathes. We'll use it only if we need it one day.

benozol:
Perfect. I’ll polish the utilities still a bit and will push to #40 later this evening.

Niols:
It's supposed to be nice, I use it in the debug logs of the constraints. By the way, if you want them to show up, you can use something like:

```ocaml
Logs.(set_reporter (Logs_fmt.reporter ()));
Logs.(set_level (Some Debug))
```

in the utility (maybe after parsing, so that we can get the log level from the command line?)

lib/constraints/src/interface.ml
```diff
@@ -0,0 +1,170 @@
...
+  (** The conjunction of two [t] (alias for {!and}). *)
+
+  val or_ : t -> t -> t
+  (** The disjunction of two [t]. *)
+
+  val var : (Var.t -> t) -> t
+  val var2 : (Var.t -> Var.t -> t) -> t
+
+  val exists : ?hint:string -> (Var.t -> t) -> t
+  val exists2 : (Var.t -> Var.t -> t) -> t
```
benozol: What about a (single) `?hint` for `exists2` as well? This would be useful for creating two corresponding variables in the old and in the new filesystem

lib/constraints/src/common/path.mli
```diff
@@ -0,0 +1,25 @@
...
+
+type t =
```
benozol: The `colis` tool has to report the CWD of symbolic execution result states. I have currently implemented it as below but it would be better part of the constraints interface, I guess.

```ocaml
let pp_comp fmt = function
  | Constraints.Path.Here -> Format.pp_print_string fmt "."
  | Up -> Format.pp_print_string fmt ".."
  | Down f -> Constraints.Feat.pp fmt f

let path_pp fmt path =
  let rel =
    match path with
    | Constraints.Path.Abs rel ->
      Format.pp_print_string fmt "/";
      rel
    | Rel rel ->
      rel
  in
  let pp_sep fmt () = pp_print_string fmt "/" in
  pp_print_list ~pp_sep pp_comp fmt rel
```

lib/constraints/src/common/var.mli
```diff
@@ -0,0 +1,17 @@
+type t
```
benozol: A function to extract the hint of a give variable would be useful to create descriptive variables

lib/constraints/src/interface.ml
Niols: I thought I had added that. it's done now

lib/constraints/src/common/path.mli
Niols: So basically, you want to be able to print pathes. I agree, it's interesting, I'll do that

lib/constraints/src/implementation/naive/constraints_implementation_naive.ml
```diff
@@ -0,0 +1,25 @@
...
+
+let eq x y = add (Pos (Eq (x, y)))
+let neq x y = add (Neg (Eq (x, y)))
+let feat x f y = add (Pos (Feat (x, f, y)))
+let nfeat x f y = add (Neg (Feat (x, f, y)))
+let abs x f = add (Pos (Abs (x, f)))
+let nabs x f = add (Neg (Abs (x, f)))
+let reg x = add (Pos (Reg x))
+let nreg x = add (Neg (Reg x))
+let dir x = add (Pos (Reg x))
```
benozol:
```suggestion
let dir x = add (Pos (Dir x))
```

lib/constraints/src/implementation/naive/constraints_implementation_naive.ml
```diff
@@ -0,0 +1,25 @@
...
+let ndir x = add (Neg (Reg x))
```
benozol:
```suggestion
let ndir x = add (Neg (Dir x))
```

lib/constraints/src/implementation/naive/constraints_implementation_naive.ml
Niols: I like how you "suggest" that when it's obviously a mistake from my end ;-)

benozol:
Getting the types right for using the continuation-style of `Clause.exists` for the intermediate root variables would be quite intrusive. Could you just expose `I.quantify_over`?

lib/constraints/src/interface.ml
```diff
@@ -0,0 +1,169 @@
...
+
+  let add_to_sat_conj = (@@)
+
+  let or_ r1 r2 = fun c ->
+    (c |> r1) @ (c |> r2)
+
+  let rec resolve x pi q z =
+    match Path.split_first_rel q with
+    | None -> eq x z
+    | Some (Down f, q) -> exists (fun y -> feat x f y & resolve y (x :: pi) q z)
```
benozol:
```suggestion
    | Some (Down f, q) -> exists ~hint:(Feature.to_string f) (fun y -> feat x f y & resolve y (x :: pi) q z)
```

lib/constraints/src/interface.ml
```diff
@@ -0,0 +1,169 @@
...
+  let resolve r cwd q z =
+    match q with
+    | Path.Abs q -> resolve r [] q z
+    | Path.Rel q -> resolve r [] Path.(rel (concat cwd q)) z
+
+  let rec noresolve x pi q =
+    (* Invariant: dir(x). *)
+    match Path.split_first_rel q with
+    | None -> (fun _ -> []) (* false *)
+    | Some (Down f, q) -> or_ (abs x f) (exists (fun y -> feat x f y & (or_ (ndir y) (dir y & noresolve y (x::pi) q))))
```
benozol:
```suggestion
    | Some (Down f, q) -> or_ (abs x f) (exists ~hint:(Feat.to_string f) (fun y -> feat x f y & (or_ (ndir y) (dir y & noresolve y (x::pi) q))))
```

benozol: These changes may prevent some of the `?₁' variables…

lib/constraints/src/interface.ml
```diff
@@ -0,0 +1,169 @@
...
+    | Some (Here, q) -> noresolve x [] q
```
benozol:
```suggestion
    | Some (Here, q) -> noresolve x pi q
```

lib/constraints/src/interface.ml
Niols: I'm not sure about the names using the features, do they really make the constraints more readable?

lib/constraints/src/interface.ml
Niols: Uh, that's weird

lib/constraints/src/interface.ml
benozol: From looking at the output of the symbolic execution, yes. I would even suggest to make the hint mandatory. You can give it a try in #40

benozol: And there is still something fishy about `noresolve`. I do not understand where the second failure state in the following comes from (constraint generation or resolution). Any idea?

```
$ ./bin/colis --run-symbolic <(echo test -e x)
Path str: x -- x
* Initial state

- root: r₁
  cwd: /
  clause: dir(r₁)

* Success states

- root: r₁₂
  cwd: /
  clause: ∃ x₃⋅ r₁ = r₁₂ ∧ r₁[x]x₃ ∧ dir(r₁)
  stdout: |
    [DBG] test -e x: path resolves

* Failure states

- root: r₁₂
  cwd: /
  clause: r₁ = r₁₂ ∧ r₁[x]↑ ∧ dir(r₁)
  stdout: |
    [DBG] test -e x: path does not resolve

- root: r₁₂
  cwd: /
  clause: ∃ x₅⋅ r₁ = r₁₂ ∧ r₁[x]x₅ ∧ dir(r₁) ∧ reg(x₅) ∧ x₅[∅]
  stdout: |
    [DBG] test -e x: path does not resolve

* Summary

- Success cases: 1
- Error cases: 2
```

Niols: I've included the modification. A mandatory hint could be a nice idea.

benozol:
(I think I added the comment below to a conversion that you were about to close, so here is it again…)

There is still something fishy about `noresolve`. I do not understand where the second failure state in the following comes from (constraint generation or resolution). Any idea?

```
$ ./bin/colis --run-symbolic <(echo test -e x)
Path str: x -- x
* Initial state

- root: r₁
  cwd: /
  clause: dir(r₁)

* Success states

- root: r₁₂
  cwd: /
  clause: ∃ x₃⋅ r₁ = r₁₂ ∧ r₁[x]x₃ ∧ dir(r₁)
  stdout: |
    [DBG] test -e x: path resolves

* Failure states

- root: r₁₂
  cwd: /
  clause: r₁ = r₁₂ ∧ r₁[x]↑ ∧ dir(r₁)
  stdout: |
    [DBG] test -e x: path does not resolve

- root: r₁₂
  cwd: /
  clause: ∃ x₅⋅ r₁ = r₁₂ ∧ r₁[x]x₅ ∧ dir(r₁) ∧ reg(x₅) ∧ x₅[∅]
  stdout: |
    [DBG] test -e x: path does not resolve

* Summary

- Success cases: 1
- Error cases: 2
```

lib/constraints/src/interface.ml
Niols: I'll fix the noresolve. I think that `noresolve` means that the path resolves **to a directory** which is not the negation of `resolve` (although, maybe `resolve` should enforce that?)

Niols: Actually, I think I also found a bug in the `resolve`...

Niols:
I still saw the notification and answered there. My `noresolve` was probably broken. But by writing it carefully I discovered a bug in `resolve`.

benozol:
Alright that looks now like what one would expect, thanks!
```
$ ./bin/colis --run-symbolic <(echo test -e x)
Path str: x -- x
* Initial state

- root: r₁
  cwd: /
  clause: dir(r₁)

* Success states

- root: r₁₂
  cwd: /
  clause: ∃ x₃⋅ r₁ = r₁₂ ∧ r₁[x]x₃ ∧ dir(r₁)
  stdout: |
    [DBG] test -e x: path resolves

* Failure states

- root: r₁₂
  cwd: /
  clause: r₁ = r₁₂ ∧ r₁[x]↑ ∧ dir(r₁)
  stdout: |
    [DBG] test -e x: path does not resolve

* Summary

- Success cases: 1
- Error cases: 1
```

benozol:
Another question: does `x[f]y` imply `dir(x)` in normalized form? This would justify the formula for the first success case below. The error case with `r₁[x]x₆ ∧ ¬dir(x₆)`, however, seems missing. Any clue?

```
$ ./bin/colis --run-symbolic <(echo 'touch x; touch x/y')
* Initial state

- root: r₁
  cwd: /
  clause: dir(r₁)

* Success states

- root: r₁₂₁₇
  cwd: /
  clause: ∃ x₆, y₂₄⋅ r₁ = r₁₂₁₇ ∧ r₁[x]x₆ ∧ x₆[y]y₂₄ ∧ dir(r₁)
  stdout: |
    [DBG] touch x: path resolves
    [DBG] touch x/y: path resolves

- root: r₁₂₁₇
  cwd: /
  clause: ∃ x₆, x₁₉, y₂₀⋅ r₁[x]x₆ ∧ r₁₂₁₇[x]x₁₉ ∧ x₁₉[y]y₂₀ ∧ x₆[y]↑ ∧ dir(r₁) ∧ dir(x₆) ∧ dir(x₁₉) ∧ reg(y₂₀) ∧ y₂₀[∅] ∧ r₁ ~{x} r₁₂₁₇ ∧ x₆ ~{y} x₁₉
  stdout: |
    [DBG] touch x: path resolves
    [DBG] touch x/y: create file

* Failure states

* Summary

- Success cases: 2
- Error cases: 0
```

Niols:
You are right, `x[f]y` implies `dir(x)`. Also, indeed, the error case that you mention is missing and I don't know why. My guess would be that it comes from an error in the resolve or noresolve, or in the specification of touch. I just had a look at `touch`s specification in #40, and I think the case where the parent resolves but isn't a directory is missing, which is exactly the case that you mentioned, am I right?

benozol:
Yes indeed! Corrected in #40.

Niols:
We need to add tests about specifications. For a specification, we should probably be able to detect that not all the preconditions are met. I'll open an issue, but I'll only do that after #40 and #41 are merged (these PRs are becoming huge)

benozol:
Currently the cases of the utilities are dynamic, i.e., they are functions that depend on a concrete argument path and make a case analysis if the path is empty and on the last component of the path.
Do you consider creating dynamic tests of the cases after generating them during symbolic execution?

Or do you suggest a declarative notion to define utilities independently from a concrete path? (This  would allow for a static analysis). For example (rough draft):

```ocaml
type specification (* declarative specification of a utility (no functional values) *)

let mkdir : specification =
  for_path
    ~empty:[error] (* mkdir '' *)
    ~last_here:[error] (* mkdir p/. *)
    ~last_up:[error] (* mkdir p/.. *)
    ~last_feature:[success …; error …; …]

(* Concrete utility with conversion from specification *)
type utility = arg:string -> state -> (bool * state) list

val apply : specification -> utility
```

Niols:
I was thinking about something simpler, like call a specification on a few well-chosen pathes (the empty path, `a/b/c`, `a/../c`, `a/b/.`, etc.) (or even random pathes) that cover most cases, get a specification for these particular pathes and run tests (that is, ask things to the prover on a few formulas).

However, I like this declarative approach pretty much. But I don't think we can be general enough to cover all cases (although if we need special treatment for only one or two utilities, it's perfectly fine…)

benozol:
Sofar (after `touch`, `mkdir`, `test -e`) the functions that create the cases only check for empty paths and the type of the trailing component. We would need to study the specifications of the remaining utilities  to see if other case analyses are required in general.

Regarding the declarative approach I am also wondering how to introduce concrete CWD, path and trailing feature (copying the specification mechanism into a language in its own with variables relevant paths may be too much effort).

benozol:
> after #40 and #41 are merged (these PRs are becoming huge)

Should we do this soon? I think we have reached a quiet stable basis and could continue development in new PRs (tests for utilities, filesystem hierarchy standard, …). If you merge #41 I’ll follow by updating the proofs in #40.

benozol:
Having spent so much time looking into this PR, I started to really appreciate the name of the branch only now-

https://xkcd.com/1296/

Niols:
No idea. Adding it does not seem to be sufficient, I did not look more into it, but it seems that the Dockerfile doesn't get the right list of dependencies from the .opam file. I'll have a look this afternoon.

Le 18 décembre 2018 11:28:26 GMT+01:00, Benedikt Becker <notifications@github.com> a écrit :
>Agreed! But travis is unhappy… do you have an idea where dune got lost?
>https://travis-ci.org/colis-anr/colis-language/jobs/469062596#L1127>
>>
>-- >
>You are receiving this because you authored the thread.>
>Reply to this email directly or view it on GitHub:>
>https://github.com/colis-anr/colis-language/pull/41#issuecomment-448172871

-- 
Envoyé de mon appareil Android avec Courriel K-9 Mail. Veuillez excuser ma brièveté.

benozol:
Wait… is the dot printer in the interface? Or how should it be used?

Niols:
Hum, you're right, there's probably something missing in the interface. Namely a function from t to Common.Conj.t so that one can use Common.Conj pp_as_dot. Or maybe I should show the function pp_as_dot directly in the interface?

Le 18 décembre 2018 11:56:33 GMT+01:00, Benedikt Becker <notifications@github.com> a écrit :
>Wait… is the dot printer in the interface? Or how should it be used?>
>>
>-- >
>You are receiving this because you authored the thread.>
>Reply to this email directly or view it on GitHub:>
>https://github.com/colis-anr/colis-language/pull/41#issuecomment-448180979

-- 
Envoyé de mon appareil Android avec Courriel K-9 Mail. Veuillez excuser ma brièveté.

benozol:
The latter would be more consistent with `Clause.pp` but I wouldn’t mind the former.

Niols:
I tried to track the CI issue down. It turns out that, for some reason, there lacks some system dependencies as well as some OCaml dependencies. However, when I try on my machine (outside Docker), everything is fine.

I assumed it was coming from `opam depext` and got rid of it by explicitely listing the required packages in `apt-get install`. And now I end up with a compilation error from Why3:
```sh
# /bin/sh: 0: Can't open ./run_autoconf_if_needed.sh
```

It's quite surprising, to say the least.

Niols:
Hum, the docker command doesn't work on my machine (and apparently not on Travis' machines either). Have you tried after cleaning your Docker cache?

Niols:
Not done yet, but I already have:

```
[...]
 ---> bf2b30ca6ac4
Step 13/27 : RUN opam show . -f depends:   | grep -A 50000 '^$'   | cut -d '"' -f 2   | xargs opam depext --install
 ---> Running in d084840b6230
# Detecting depexts using vars: arch=x86_64, os=linux, os-distribution=debian, os-family=debian
# No extra OS packages requirements found.
# All required OS packages found.
Removing intermediate container d084840b6230
 ---> dd84d9e4fbd7
Step 14/27 : WORKDIR /home/opam/provers
[...]
```

which means that the OPAM dependencies were not installed. That cannot be good.

Niols:
By the way, I just noticed that the compilation of external provers happens **after** the `COPY` statement in the Dockerfile. This is not such a good idea, because it means that even the smallest change in the file tree invalidates the cache including the compilation of the provers.

The same goes for the installation of dependencies. It's better to just include the necessary files (like the `.opam` file only), to install the dependencies and then to have the full `COPY . .`.

Niols:
> Not done yet, but I already have:

It's done now, and it failed complaining that Dune doesn't exist.

Niols:
It's because `RUN opam show . -f depends:   | grep -A 50000 '^$'   | cut -d '"' -f 2` doesn't print *a thing*. I don't know why. It does on my computer…

Niols:
OK, OPAM 2.0.2 broke a few things. I changed the Dockerfile. But then, Why3 complained. I tried the last dev commit but the option `--add-prover` didn't seem to be correctly handled. I then took the first version of Why3 correctly handling OPAM 2 and it went rather well, except that the session is obsolete, and that the replay fails

benozol:
Thanks @niols for tracking this down.

We could try now either to fix the proof sessions in this PR to merge it on master, or to
rebase #40 on this branch (done) and fix the proofs there before merging all at once, finally.
I hope but cannot promise that could happen still this year. WDYT?

Niols:
Hum, actually, I don't mind. I usually try to keep master compiling / passing the CI. But master is sooo far behind that it's really not important. Since you're the one that is going to fix the proofs, choose whatever is more convenient to you!

Niols:
Agreed! :-)

PR #47, opened 2019-01-16
Add utility cd

### LINKED ISSUE #33: Add builtin  `cd`

Niols:
fix #33 

I currently added it to the syntax, to the pretty printer and to the converter from Shell. We still have to add it to the semantics and the interpreter.

```
File "src/symbolic/symbolicInterpreter.mlw", line 186, characters 4-7241:
warning: Non-exhaustive pattern matching, asserting `absurd'
File "src/concrete/interpreter.mlw", line 179, characters 15-5155:
warning: Non-exhaustive pattern matching, asserting `absurd'
```

Comments?


Niols:
Note: obviously, CI fails on this branch

Niols:
Also, we need to add this to the CoLiS parser!

benozol:
The agreement was to implement `cd` as a "special builtin", i.e., in the interpreter and not as a utility, right? I’d add a commit here then.

Niols:
Yes. Although this is not "officially" a special built-in for historical reasons, it still has to change something in the states, so I would have it added like a component of the language like I just did. However, an other way might be possible.

Niols:
s/I just did/I just started/

benozol:
Awe and terror: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/cd.html

What about the following simplification for CoLiS:

- OCaml’s `Sys.chdir` for concrete interpretation (with or without `--real-world`)
- resolve argument as directory and set `state.filesystem.cwd` in the success casses for symbolic interpretation

---

BTW The disadvantage of implementing `cd` in the symbolic interpreter is that it would require the exposure of `resolve` to the symbolic engine, which is currently nicely reduced and non-functional:

https://github.com/colis-anr/colis-language/blob/1571a69572799a2ef9ef822625812981179ab7af/src/symbolic/driver.drv#L1-L6

Implementing it as a utility is currently straightforward because utilities can modify the symbolic state that contains the cwd. **However**, this might be a mistake in its own and we should move the cwd out of the filesystem:

```diff
--- type filesystem = { root ; clause ; cwd }
--- type state = { stdout ; stdin ; filesystem }
+++ type filesystem = { root ; clause  }
+++ type state = { stdout ; stdin ; filesystem ; cwd }
```

and reduce the signature of utilities to the bare minimum

```diff
--- type utility = state → (state × bool) list
+++ type result = { clause: sat_conj ; output: stdout ; messages: stdout ; result: bool }
+++ type utility = cwd:path → clause:sat_conj → result list
```

Any thoughts?

benozol:
`cd` is really a very special builtin.

The symbolic engine differentiates between the *state*, which describes the state of your computer, including stdout and filesystem, and the *context*, which corresponds to the language interpreter context. Only the state is shared with and can be changed by subprocesses and symbolic utilities, one interpreter’s context can notably not be modified by subprocesses or utilities. I think this is the proper way, but it means that `cd` cannot be a utility, because its interpretation requires changing the variable environenment (`PWD`) and the field of `context.cwd` that contains the same info as a `Path.normal`.

On the other side, interpreting `cd` introduces constraints and variables (`resolve`), which is (sofar) not intended in the symbolic engine.

I think we have to implement the constraint generation of `cd` in OCaml, expose this to Why3, and change the context in the symbolic engine. How many cases of `cd` were there in the corpus? :)

benozol:
This PR implements `cd` as a *special builtin*, that is, in the symbolic engine, and using the `test -d` utility to test if the path (prefixed with the current PWD if relative) exists.

src/concrete/interpreter.mlw
```diff
@@ -263,6 +263,7 @@ module Interpreter
     | ISubshell ins ->
       let sta' = {
         sta with
+        cwd = ref !(sta.cwd);
```
Niols: I don't get why it has to be a reference?

Niols:
We have a problem with CI:

```
The following actions will be performed:
  - downgrade camlzip        1.08 to 1.07 [required by alt-ergo]
  - install   conf-autoconf  0.1          [required by alt-ergo]
  - recompile why3           dev*         [uses camlzip]
  - install   psmt2-frontend 0.1          [required by alt-ergo]
  - install   ocplib-simplex 0.4          [required by alt-ergo]
  - install   alt-ergo       2.2.0
```

but I don't think it is related with this PR. Most likely more with version constraints on `calmzip`.

There aren't many occurrences of `cd` in the corpus. But it's an emblematic utility; hard not to have it, I guess.

Niols:
Oh but you fixed that already.

benozol:
> Oh but you fixed that already.

There seems to have been a release of camlzip (1.08) in early july that the CI. I don’t understand the problem really, e0f38ce is only a workaround.

src/concrete/interpreter.mlw
benozol: Well, it can now be changed using `cd` :) The reason for using a reference instead of a mutable field is exactly for subshells, where the subshell receives only a copy of the reference and is allowed to change that, but leaves the caller’s cwd unchanged (same for `var_env`, and `arguments` in function calls)

benozol:
Yes indeed, no clue what it means without. BTW the comment was also wrong, it’s alt-ergo 2.2.0 that requires `camlzip < 1.08`

benozol:
@niols Apparently these lights won’t turn much greener (the CI finished successfully), so I’ll merge if you don’t have any further remarks

Niols:
Oh, yes, indeed, feel free to merge.

PR #48, opened 2019-01-17
harmonize handling of return codes

### LINKED ISSUE #32: Decide the meaning of the return codes, specify it and update the code accordingly

Niols:
fix #32 

This PR starts harmonizing and documenting the return codes used by our tool.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,3 +30,20 @@ output:
   stderr: <a string>
   return_code: <an integer>
 ```
+
+## Return Codes
+
+The tool uses different return codes for different kind of errors.
+
+| Code | Meaning
+|------|---------
+|   0  | Success
+|   1  | Success of the concrete execution ending up on a failure
+|   2  | Unhandled/unexpected OCaml exception
+|   3  | Error in command-line parsing
+|   4  | Error while reading input file
+|   5  | Error in parsing (Shell or CoLiS)
+|   6  | Error in conversion
+|   7  | Unsupported utility
+|   8  | Unsupported option for a supported utility
+|   9  | Error in pretty-printing (Shell or CoLiS)
--- a/src/colis.ml
+++ b/src/colis.ml
@@ -33,6 +33,7 @@ end
 
 (* CoLiS *)
 
+exception FileError of string
 exception ParseError of string * Lexing.position
 exception ConversionError of string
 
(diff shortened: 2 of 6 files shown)


benozol:
Looks good sofar.

But failing on any unknown utilities may not correspond to normal shell behaviour where this can be handled (e.g., `guga || true`). I am wondering if we should

1. make the behaviour configurable (by a command line option), and 
2. implement a corresponding behaviour for the concrete interpreter.

Niols:
This could (should?) indeed be an option of the tool. I can try to add that and come back to you. The same option could also apply to the concrete interpretation indeed.

How would you name this option? Because `strict` is already taken. Maybe a long name like `--fail-on-unknown-utility`?

Niols:
Also, I want your opinion on this. Such options (the same applies to `--realworld`) are not having an impact on the interpreters really, but rather on the implementation of the utilities. We basically have two choices:

1. We can have an option passed from the Colis_cmd module to the interpreters that pass them again to their respective implementation of utilities.

2. We can have global references somewhere that we update according to what is passed in the command line.

The former is cleaner OCaml. However, the latter allows to not bother about these arguments in the interpreters which makes their code cleaner and avoids changing them every time we want to add something new.

I would personally go for the latter. Any opinion?

benozol:
I wouldn’t mind global variables for global options at this point either.

A third option would be to configure the behaviour by environment variables that are read where needed, but it would be weird to mix environment variables and command line options.

Niols:
Hum. Actually, we could use environment variables. But I think that they should be handled at the same time as the command-line parsing (with one having the priority over the other---command-line parsing probably)

Niols:
What do you think of that last commit?

src/concrete/utilities.ml
```diff
@@ -8,44 +8,47 @@ open Semantics__Buffers
...
+        in
+        let sta' = {sta with stdout; stdin=Stdin.empty} in
+        sta', result
+     | _ ->
+        let str = "grep: not exactly one argument" in
+        let stdout = Stdout.(sta.stdout |> output str |> newline) in
+        {sta with stdout}, false
+     end
+  | _ ->
+     if !Options.fail_on_unknown_utilities then
```
benozol: Can’t we use `unknown_utility` here?

benozol: Well, the concrete version of it but I see there is none…

Niols: I can add one if you think it makes the code clearer

Niols:
The only exception I didn't move is ColisLexer.LexerError. I'm not so sure about this choice. I thought to myself that this was meant to be an internal error that the Colis API does not expose so it would make sense not to have it in Errors. wdyt?

benozol:
Yes agreed, lexing errors are just converted to and reported as parsing errors.

benozol:
And if the CI approves this PR, I will do so too.

Niols:
There is one commit that failed because of a timeout while replaying the proofs. However, I don't see how my PR changes that…

PR #49, opened 2019-01-17
Move lexer and parser back where they belong

### LINKED ISSUE #10: Move colisLexer and colisParser back where they belong
This was blocked by an issue in Dune (https://github.com/ocaml/dune/issues/1372) which is now fixed. However, fixing it in this project would mean that we require the last version of Dune (and probably the development version). This issue is here to remind us that some day later, we should fix it.

Niols:
fix #10 

I don't think there is much to say here. Dune 1.6 seems to have reached testing (https://packages.debian.org/testing/ocaml/) and the issue that blocked use was fixed in Dune 1.4, so I guess we can close this.

@benozol If that's OK with you, feel free to just merge this.


benozol:
OK that’s nice.

PR #52, opened 2019-01-24
Complete symbolic interpretation

### LINKED ISSUE #51: Complete symbolic execution
Currently, the symoblic interpretation of arguments and substrings is not supported.

benozol:
Fixes #51

### DIFF
--- a/src/symbolic/driver.drv
+++ b/src/symbolic/driver.drv
@@ -26,6 +26,7 @@ module symbolicAuxiliaries.Fset
   syntax val map "BatSet.map %1 %2"
   syntax val fold "BatSet.fold %1 %2 %3"
   syntax val filter "BatSet.filter %1 %2"
+  syntax val bind "BatSet.fold BatSet.union (BatSet.map %1 %2) BatSet.empty"
   syntax val partition "BatSet.partition %1 %2"
   syntax val to_list "BatSet.elements %1"
   syntax val of_list "BatSet.of_list %1"
--- a/src/symbolic/symbolicAuxiliaries.mlw
+++ b/src/symbolic/symbolicAuxiliaries.mlw
@@ -49,6 +49,9 @@ module Fset
   val function map (f: 'a -> 'b) (s: set 'a) : set 'b
     ensures { result = map f s }
 
+  val function bind (f: 'a -> set 'b) (s: set 'a) : set 'b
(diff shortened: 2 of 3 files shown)


Niols:
The CI does not seem to be happy. But I guess it's the reason for the 'wip' label.

I've got a question: Why use BatSet? From what I can see, you're only using the part that is already in Set, am I right?

benozol:
A typo broke the CI (fixed). 

I marked the PR as WIP because the added code is quite complex (e.g., [three levels of recursion for for-loops](https://github.com/colis-anr/colis-language/blob/6d05242e8f341e010e33f35f1cfbaed601ef7016/src/symbolic/symbolicInterpreter.mlw#L325-L362)) and may deserve some additional checks. Should we think about testing the symbolic interpreter? Prooving the approximation properties may take some time for CoLiS…

> Why use BatSet?

For their polymorphic sets. Instantiating `Set.Make` in Why3 for all the different element types currently requires considerable amount of boilerplate (drivers for each instantition!). To replace the use of polymorphic BatSets, I’d prefer to wait for the integration of OCaml’s standard containers into the Why3 API, which had been discussed by the team.

Niols:
> A typo broke the CI (fixed).
> 
> I marked the PR as WIP because the added code is quite complex (e.g., [three levels of recursion for for-loops](https://github.com/colis-anr/colis-language/blob/6d05242e8f341e010e33f35f1cfbaed601ef7016/src/symbolic/symbolicInterpreter.mlw#L325-L362)) and may deserve some additional checks. Should we think about testing the symbolic interpreter? Prooving the approximation properties may take some time for CoLiS…

We could and should definitely think about testing the symbolic interpreter. And we had various ideas to do that.

> > Why use BatSet?
> 
> For their polymorphic sets. Instantiating `Set.Make` in Why3 for all the different element types currently requires considerable amount of boilerplate (drivers for each instantition!). To replace the use of polymorphic BatSets, I’d prefer to wait for the integration of OCaml’s standard containers into the Why3 API, which had been discussed by the team.

OK, it makes sense. Where does the comparison function come from?

benozol:
For the moment it’s just `BatSet.empty` with `Pervasives.compare`. But we could implement proper comparison functions and use `BatSet.create` instead. Is there a comparison for `sat_conj` in the constraints interface?

benozol:
Superseded by #60

PR #54, opened 2019-01-24
Reorganize Colis.run_symbolic and return false for any failures

### LINKED ISSUE #53: Symbolic execution return codes

benozol:
Fixes #53

### DIFF
--- a/src/colis.ml
+++ b/src/colis.ml
@@ -124,20 +124,15 @@ let print_symbolic_filesystem fmt fs =
   fprintf fmt "cwd: %a@\n" Path.pp fs.cwd;
   fprintf fmt "clause: %a@\n" Clause.pp_sat_conj fs.clause
 
-let _print_dot filename id clause =
-  let ch = open_out filename in
-  try
-    let fmt = formatter_of_out_channel ch in
-    Constraints.Conj.pp_as_dot ~name:id fmt clause;
-    close_out ch
-  with e ->
-    close_out ch;
-    raise e
-
-let print_symbolic_state _id fmt sta =
+let print_symbolic_state fmt ?id sta =
   let open Symbolic.State in
+  begin match id with
+    | Some id ->
+      fprintf fmt "id: %s@\n" id;
+      (* print_dot (sprintf "/tmp/%s.dot" id) id sta.filesystem.clause; *)
+    | None -> ()
+  end;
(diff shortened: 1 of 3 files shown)


src/colis.ml
```diff
@@ -124,20 +124,15 @@ let print_symbolic_filesystem fmt fs =
   fprintf fmt "cwd: %a@\n" Path.pp fs.cwd;
   fprintf fmt "clause: %a@\n" Clause.pp_sat_conj fs.clause
 
-let _print_dot filename id clause =
-  let ch = open_out filename in
-  try
-    let fmt = formatter_of_out_channel ch in
-    Constraints.Conj.pp_as_dot ~name:id fmt clause;
```
benozol: @Niols could you expose `pp_as_dot` also for `sat_conj`?

benozol:
CI fails due to a timeout in replaying the proofs in 1 of 6 containers :(

Niols:
Can't we extend the timeout?

src/colis.ml
```diff
@@ -154,40 +149,47 @@ let print_symbolic_state _id fmt sta =
 
 let run_symbolic ~argument0 ?(arguments=[]) fs_spec colis =
   let open Symbolic in
-  let states : State.state list =
+  let prune_init_state = false in
```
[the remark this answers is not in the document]
Niols: What the difference? An existential quantifier over the first root? This could indeed be a command-line flag (now that we have global references for options, it's rather easy) that we could enable to make the execution (much) faster.

src/colis.ml
Niols: It is already the case!

benozol: Could you remind me where it is?

src/colis.ml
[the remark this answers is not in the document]
benozol: Here we go: the functionality was already there without global state, and I reused that.

benozol:
Good enough? We can leave integrating the dot printer to another PR.

Niols:
I think it is. Let's have an issue assigned to me, this dot printer for sat_conj has waited long enough.

For the moment, Id say we can merge this (and I'll re-run colis-batch).

src/colis.ml
Niols: Actually, no, it's not exposed… I'll handle that in an other PR.

PR #57, opened 2019-01-24
remove morsmall from pin-depends

Niols:
We have to wait for the OPAM image in DockerHub to be up to date with the repositories. After that, we will be able to trigger the CI on this pull request and merge it into master.


Niols:
Hum. Finally, Morsmall is moving a little so it might not be the best idea to depend on a version here.

PR #60, opened 2019-01-25
Bounded while loops

benozol:


Niols:
4/4, just waiting for the CI?

benozol:
Yes, fingers crossed

benozol:
BTW the scripts for testing the symbolic execution are not yet included in the dune tests because a) there is no infrastructure, and because b) I don’t know how to write a terminating while loop in CoLiS at the moment oO

benozol:
If we only had such advanced functionality like string equality tests...

Am 29. Januar 2019 02:26:52 MEZ schrieb Nicolas Jeannerod <notifications@github.com>:

Niols:
That'll come soon. The parser for test expressions is being integrated in Morsmall as we speak (colis-anr/morsmall#8). After that, adding the test utility will be very easy. But maybe in an other PR?

src/colis.ml
```diff
@@ -113,7 +113,11 @@ let run ~argument0 ?(arguments=[]) colis =
   let input = { Input.empty with argument0 } in
   let state = Interpreter.empty_state () in
   state.arguments := arguments;
-  Interpreter.interp_program input state colis;
+  let cnf = Z.minus_one in
```
Niols: Why "cnf"?

Niols: Also, why Z and not int? In order to make Why3's goals simpler?

src/colis.ml
```diff
@@ -187,18 +191,30 @@ let run_symbolic ~prune_init_state fs_spec ~argument0 ?(arguments=[]) colis =
     fprintf fmt "- @[%a@]@\n" (print_symbolic_state ~id) sta
   in
   List.iter
-    (fun (init, normals, errors) ->
+    (fun (init, (normals, errors, failures)) ->
```
Niols: I was expecting an exception to be raised. But in fact, you're creating a new behaviour that corresponds to the failures? Why so?

Niols:
What I read looked good to me :-)

src/colis.ml
benozol: "cnf" configuration of the symbolic execution, a Why3 record with a single integer field representing the loop limit, that is unboxed to a single Z.t on extraction. But yes, the code is misleading and I’ll clarify.

Niols:
Oh. Is Why3 always unboxing records with only one field? That seems to be a bit error-prone for users of the extraction mechanism.

Maybe a simple comment would be enough.

(And tbh I first though of "conjunctive normal form")

src/colis.ml
benozol: For two reasons:

- the failure behaviour is also included in the semantics, which allows for proving the over-approximation property (“symbolic completeness”)

- we can identify execution paths that yield to errors (i.e. positive return value) in the presence of failures in the same program (e.g., https://github.com/colis-anr/colis-language/blob/e285ac87f6145ec21f74c38c2d2e4695c4f26348/tests/symbolic/error-failure.sh), and unimplemented options and utilities could take advantage of this, too.

The drawback is of course the inclusion of the failure behaviour in the semantics, whithout any correspondance in POSIX Shell. It may be worth to show that the failure behaviour is not triggered in concrete execution with a negative loop limit.

benozol:
(We can try if `why3 replay --use-steps` helps. I am not sure if all provers we use support this option, though.)

src/colis.ml
Niols: > It may be worth to show that the failure behaviour is not triggered in concrete execution with a negative loop limit.

Indeed, but it is probably not worth having that as a first priority. I could be stated somewhere though.

benozol:
> you're creating a new behaviour that corresponds to the failures? Why so?

(@Niols @claudemarche I think this point requires more discussion)

The behaviour ("`failure`") that results from hitting the iteration limit in while loops is on same level as the normal, exit, and return behaviours. This has two advantages:

1. We can identify execution paths that yield to errors (i.e. positive return value) in the presence of failures in the same program (e.g., https://github.com/colis-anr/colis-language/blob/e285ac87f6145ec21f74c38c2d2e4695c4f26348/tests/symbolic/error-failure.sh). Unimplemented options and utilities could take advantage of this, too.

2. Modelling the failure semantics in the semantics gives us certainty of catching all bugs below the loop limit, if we prove over-approximation (“symbolic completeness”).

This PR also prepares point 2: It models the failure behaviour for while loops in the semantics, imposes the loop limit in the concrete interpreter, and proves its soundness. (The loop limit can be disabled in the concrete interpreter.)

The failure behaviour, however, complicates the semantics, whithout correspondance to POSIX Shell. If we say that over-approximation and the certainty to catch all bugs below the loop limit is not our focus, we could remove the failure behaviour from the semantics and interpreter. What do you think?

Niols:
I find the way you did it nice, I was just a bit surprised at first. I think that this choice should be documented somewhere. We should also state the fact that the concrete interpreter can never reach failure states (and even prove this fact one day maybe.)

benozol:
(It seems that `why3 replay --use-steps` does not end ever. Bug in Why3?)

Niols:
It seems very good to me!

PR #63, opened 2019-01-28
conversion of the case construct

Niols:


Niols:
The CI will break. For some reason, the colis script:

```
function f begin
  case_var := arg 1
  if
    if
      if false then
        true
      else
        test [ case_var; '='; 'upgrade' ]
      fi
    then
      true
    else
      test [ case_var; '='; 'update' ]
    fi
  then
    echo [ 'foo' ]
  else
    if
      if false then
        true
      else
        test [ case_var; '='; 'clean' ]
      fi
    then
      echo [ 'bar' ]
    else
      echo [ 'baz' ]
    fi
  fi
end

begin
  call f [ 'upgrade' ]
  call f [ 'clean' ]
  call f [ 'cleana' ]
  call f [ 'shproutz' ]
end
```

isn't well parsed and I couldn't understand why. The first "begin" does not seem to be recognized as a keyword and thus the first line is understood as the call to the utility "function" with arguments "f" and "begin"… Then everything goes wrong.

I don't really know why. From what I read (but I did read fast), the lexer and the parser are correct

Niols:
Nevermind, it was my bad. I would deserve a commit message "can't fix stupid". I'll wait for the CI and merge if it passes.

PR #72, opened 2019-02-04
Command line flag to print symbolic states as dot files

benozol:


lib/constraints/src/common/conj.ml
```diff
@@ -58,7 +58,7 @@ let pp_as_dot ~name fmt (es, c) =
 
   let hash = Hashtbl.hash in
 
-  fpf fmt "@[<h 2>digraph %s {@\n" name;
+  fpf fmt "@[<h 2>digraph %S {@\n" name;
```
Niols: What does it change?

src/colis_cmd.ml
```diff
@@ -69,14 +69,15 @@ let speclist =
     "--symbolic-fs",               String set_symbolic_fs,        " Name of the initial symbolic filesystem in symbolic execution (values: empty, simple, fhs, default: empty)";
```
Niols: Shouldn't that be `"FS Name of the ..."`?

lib/constraints/src/common/conj.ml
benozol: S: convert a string argument to OCaml syntax (double quotes, escapes).

Graphviz wasn’t happy with the dashes in the graph name.

src/colis_cmd.ml
benozol: Alright!

lib/constraints/src/common/conj.ml
Niols: Oh, I didn't know that. Nice :-)

benozol:
Good to go?

Niols:
Yes, I think so.

PR #76, opened 2019-02-07
Make interp_utility take the environment

### LINKED ISSUE #73: Add utility `which`
Hey @benozol 

I'm thinking of adding the `which` utility. It's no hard work *per se*, but I need to know how to proceed. I guess it is a bit related to #33 in a way. At some point, we will have to implement utilities that depend on the state of the environment variables. This implies several questions:

- How do we want to represent utilities that read the environment? Should the function `SymbolicUtilities.interp` take the environment as well? And maybe give it only to the utilities that require it?

- Do we model the "export" mechanism of the Shell? "export" is actually [the number 1 target of the conversion now](http://ginette.informatique.univ-paris-diderot.fr/~niols/colis-covering-report/#conversion-error) and it would probably be more accurate.

- In #33, you suggested that `cd` could be implemented as a utility changing the environment. How would we model that? But maybe that's more a discussion for #33.

Niols:
fix #73


src/symbolic/utilitiesSpecification.ml
```diff
@@ -28,33 +29,32 @@ type case = {
...
+     spec = Clause.true_ }]
 
 let quantify_over_intermediate_root state conj =
   if BatOption.eq ~eq:Var.equal state.filesystem.root0 (Some state.filesystem.root) then
     [conj]
   else
     Clause.quantify_over state.filesystem.root conj
 
+let apply_output_to_state state str =
+  { state with stdout = Stdout.(output str state.stdout ) }
```
benozol: 1. I think you will be missing a newline in `state.stdout`. Shouldn’t `case.output` also be an option just as `error_message`?
2. Is this the same function as `Symbolic_specifications.print_line`?

src/symbolic/utilitiesSpecification.ml
```diff
@@ -28,33 +29,32 @@ type case = {
...
-    descr = "";
-    error_message;
-    spec = Clause.(exists @@ fun v -> eq v v); (* is there a better way to write `true' ? *)
-  }]
+let failure ?error_message () =
+  [{ result = false ;
+     descr = "" ;
+     output = "" ;
+     error_message ;
+     spec = Clause.true_ }]
```
benozol: Finally true :)

src/symbolic/symbolicInterpreter.mlw
```diff
@@ -204,7 +204,7 @@ module Interpreter
   use Results as Rs
 
   (* Implemented in OCaml *)
-  val function interp_utility identifier (list string) state : set (state, bool)
+  val function interp_utility identifier var_env (list string) state : set (state, bool)
```
benozol: Good thing that all our variables are `export`!

src/symbolic/utilitiesSpecification.ml
benozol: @Niols Do you agree? Should I give it a try? This is blocking @claudemarche ’s progress in #73

Niols: Hey. I don't know about the `newline`. I wanted to let the choice to the user: if there is a newline, it must be written in the spec. If not, then I don't want `apply_output_to_state` to add it (wrongly) for me. Is that a problem?

Niols: I assume that it's the same thing as `Symbolic_specifications.print_line` without a final newline?

Niols:
Alright, the CI is happy now. There is only this newline question remaining.

src/symbolic/utilitiesSpecification.ml
benozol: > … it must be written in the spec. … Is that a problem?

You mean by including the `\n` in `output`? Yes, that would be a problem because `state.stdout` is a buffer, i.e. supposed to be a list of _lines_. A newline within a line is … strange and will be incompatible with `read` (once we implement it). On the other hand, not calling `newline` here results in utilities terminating at the end of the line and not in a new line, which is quite uncommon.

> I wanted to let the choice to the user

Changing the type of `output` to `Stdout.t` would be then the solution (printing several lines, or ending mid-line). You can merge this into the state by using `Stdout.concat`.

Niols: Can't we just `split_on_char '\n'` and add all this to the buffer? Can this buffer represent a stdout that doesn't end on `\n`? I remember us talking about having the semantics of the Buffer of lines be that of `String.concat "\n"`, is that how it is implemented?

benozol: Yes, that’s how it is implemented. And, sure, `split_on_char '\n'` is also possible. It just seems strange to me to split here just to concat there.

BTW Stdout has very composable printers: `out |> output "abc" |> newline |> output "def |> newline`, or even `ksprintf output "hello %s" you`. From that point-of-view, the easiest type for `output` may be `stdout → stdout`.

But if you prefer to keep the interface of `UtilitiesSpecification` free from buffers, let’s go with split&concat.

Niols: No, I haven't made my mind yet, so I'll go with whichever looks best to you. `stdout → stdout`?

benozol: `?(output:Stdout.t=Stdout.empty)` using `Stdout.concat` may be most straightforward for the interface, apart from `?(output="")` using `split`.

Niols: Like so?

Niols:
Hum, I'm getting rather complicated conflicts, including at some places where I didn't change anything… Weird.

Niols:
I remember reading something about Why3 wrongly adding a unit argument to function calls. This is exactly the bug that I get after merging master into which. I'm currently pinned at 41d0c60df, is that the source of the problem?

benozol:
Yes, this bug was fixed for the Why3 version 1.2 which is currently being released. Why3 is pinned on commit `c0758c6b` in colis-language, and I removed the workarounds ec4c5dd7265b6be97f5bc9c9c8fe7f6ae2a770c7.

claudemarche:
Now that the support of `which` is in master, you may want to rerun the ginette bench.

Beware! You must make sure that symbolic interpreter is called with --symbolic-fs simple

PR #79, opened 2019-02-11
Proove termination of symbolic execution

benozol:
This PR 
1. defines an optional, maximal stack height in the semantics; only function calls increase the stack (not subshells, not utilities (yet)). 
2. The proof of the soundness of the concrete interpreter is updated by requiring the absence of a stack height (as it is for the loop limit).
3. The termination of the symbolic interpreter is proven by requiring a positive stack height (again, like the loop limit) and by a skeleton of CoLiS instructions as variants.


Niols:
> This PR
> 
>     1. defines an optional, maximal stack height in the semantics; only function calls increase the stack (not subshells, not utilities (yet)).

Subshells do not cause problems for the termination, do they?

>     2. The proof of the soundness of the concrete interpreter is updated by requiring the absence of a stack height (as it is for the loop limit).
> 
>     3. The termination of the symbolic interpreter is proven by requiring a positive stack height (again, like the loop limit) and by a skeleton of CoLiS instructions as variants.

That is, you have a lexicographic order of some sort with the stack limit > the loop limit > a reflection of the derivation tree of the semantic judgment? Are these "skeletons" the same as the ones I used in the previous iteration of the language? Or is there a way to have reflection in Why3 now? Or something else?

Makefile
```diff
@@ -1,10 +1,10 @@
-
 # Targets to replay proofs for the concrete interpreter
 replay-concrete-proofs=$(patsubst %, replay-concrete-proof-%, auxiliaries semantics interpreter)
+replay-symbolic-proofs=$(patsubst %, replay-symbolic-proof-%, symbolicInterpreter)
 
 .PHONY: ci build test doc clean install uninstall \
-  extract-why3 clean-why3 \
-  replay-proofs $(replay-concrete-proofs)
+  extract-why3 clean-why3 replay-proofs \
+  $(replay-concrete-proofs) $(replay-symbolic-proofs) \
```
Niols: Doesn't that hurt to end on a line continuation?

Makefile
```diff
@@ -55,9 +55,14 @@ clean-why3:
 
 ## Replay Why3 proofs
 
-replay-proofs: $(replay-concrete-proofs)
+replay-proofs: $(replay-concrete-proofs) $(replay-symbolic-proofs)
 
 $(replay-concrete-proofs): replay-concrete-proof-%: src/concrete/%.mlw src/concrete/%/why3session.xml
-	why3 replay \
+	why3 replay --use-steps \
```
Niols: Last thing I remember about `--use-steps` is that it didn't seem to change anything. Does it?

src/colis.ml
```diff
@@ -149,7 +149,7 @@ let print_symbolic_state fmt ?id sta =
     fprintf fmt "  %s" sta.stdout.line
   end
 
-let run_symbolic ~prune_init_state ~loop_limit ~fs_spec ~argument0 ?(arguments=[]) colis =
+let run_symbolic ~prune_init_state ~loop_limit ~stack_size ~fs_spec ~argument0 ?(arguments=[]) colis =
```
Niols: `prune_init_state`, `loop_limit` and now `stack_size`… This is starting to make a lot of arguments. Do you think it would be better to merge them in a `config` argument?

Makefile
benozol: Yes, Claude fixed the problem with `--use-steps` by now, and this PR uses the latest master of Why3

src/colis.ml
```diff
@@ -162,7 +162,9 @@ let run_symbolic ~prune_init_state ~loop_limit ~fs_spec ~argument0 ?(arguments=[
   let run_in_state sta =
     let inp = { Concrete.Input.empty with argument0 } in
     let ctx = { Context.empty_context with arguments } in
-    sta, Interpreter.interp_program (Z.of_int loop_limit) inp ctx sta colis
+    let loop_limit = Z.of_int loop_limit in
+    let stack_size = Z.of_int stack_size in
+    sta, Interpreter.interp_program loop_limit stack_size inp ctx sta colis
```
Niols: So the Why3 code works with unbounded integers right? But the OCaml interface uses `int`? That's the right way to go imo, but I just wanted to be sure I understood the code correctly.

Makefile
benozol: It’s horrendous.

src/colis_cmd.ml
```diff
@@ -69,6 +70,7 @@ let speclist =
     "--symbolic-fs",               String set_symbolic_fs,        "FS Name of the initial symbolic filesystem in symbolic execution (values: empty, simple, fhs, default: empty)";
     "--prune-init-state",          Set prune_init_state,          " Prune the initial state in symbolic execution";
     "--loop-limit",                Int ((:=) loop_limit),         sprintf "LIMIT Set limit for symbolic execution of while loops to LIMIT (default: %d)" !loop_limit;
+    "--stack-size",                Int ((:=) stack_size),         sprintf "SIZE Set the stack size for symbolic execution to LIMIT (default: %d)" !stack_size;
```
Niols: to SIZE and not to LIMIT

src/symbolic/driver.drv
```diff
@@ -12,9 +12,8 @@ end
 module symbolicInterpreter.Env
   syntax type t "(string, %1) BatMap.t"
   syntax val empty "BatMap.empty"
-  syntax val get "BatMap.find_default %3 %2 %1"
+  syntax val get "BatMap.find %2 %1"
```
Niols: Hum, why ths change?

src/colis.ml
benozol: Yes, a Why3 `int` is exported as `Z.t`, and I tried to keep that again out of the interface of module `Colis`.

Niols: Nice

benozol:
> Subshells do not cause problems for the termination, do they?

No, because their evaluation reduces the evaluated instruction (“`ISubshell ins > ins`”)

> That is, you have a lexicographic order of some sort with the stack limit > the loop limit > a reflection of the derivation tree of the semantic judgment?

Exactly.

> Are these "skeletons" the same as the ones I used in the previous iteration of the language? Or is there a way to have reflection in Why3 now? Or something else?

I guess so they are the same that you used: a skeleton is just the shape of a instruction, including list expressions and string expressions. To my knowledge this is the (only?) way to provide variants for functions on mutual, nested datatypes in Why3.

Niols:
> > Are these "skeletons" the same as the ones I used in the previous iteration of the language? Or is there a way to have reflection in Why3 now? Or something else?
> 
> I guess so they are the same that you used: a skeleton is just the shape of a instruction, including list expressions and string expressions. To my knowledge this is the (only?) way to provide variants for functions on mutual, nested datatypes in Why3.

Oh, so actually they might not be the same. You are using the CoLiS AST value as the variant, right? Which you can do because it always decreases *except* in places where you have a limit that decreases. Nice.

src/symbolic/driver.drv
benozol: `Env.t` was a `(id, α) map` before (and required a default for `get`) and is now a `(id, α option) map`, so the default is “built-in”.

BTW in  branch #69, SymbolicInterpreter.Env is replaced by Semantics.Env.

benozol:
> You are using the CoLiS AST value as the variant, right? 

Not sure if I understand correctly, but the skeletons are used as variants, it’s their sole purpose so far (`let interp_instruction … ins variant { stack, loop, skeleton ins }`).

src/colis.ml
benozol: Agreed.

Niols:
I'm happy with this PR;

PR #80, opened 2019-02-11
Command line arguments

benozol:
This PR allows (for concrete and symbolic interpreter)
- giving arguments to the interpreter (e.g. for `$1`) even when they start with a dash (they were matched with arguments to the command line tool before)
- defining variables from the command line. Example

```shell
$ cat test.sh
echo $0: $1/$2/$a/$b
$ ./bin/colis --var a=X --var b=B --var a=A test.sh one -- --two
test.sh: one/--two/A/B
```

### DIFF
--- a/src/colis.ml
+++ b/src/colis.ml
@@ -98,11 +98,18 @@ let colis_to_file filename colis =
 
 (* Interpret *)
 
-let run ~argument0 ?(arguments=[]) colis =
+let mk_concrete_var_env =
+  let open Concrete.Env in
+  List.fold_left
+    (fun env (var, val_) -> mixfix_lblsmnrb env var val_)
+    (empty_env "")
+
+let run ~argument0 ?(arguments=[]) ?(vars=[]) colis =
   let open Concrete in
   let input = { Input.empty with argument0 } in
   let state = Interpreter.empty_state () in
   state.arguments := arguments;
+  state.var_env := mk_concrete_var_env vars;
   Interpreter.interp_program input state colis;
   print_string (Stdout.all_lines !(state.stdout) |> List.rev |> String.concat "\n");
   exit (if !(state.result) then 0 else 1)
@@ -156,7 +163,10 @@ type symbolic_config = {
   stack_size: int;
 }
 
-let run_symbolic config ~argument0 ?(arguments=[]) colis =
+let mk_symbolic_var_env vars =
+  BatMap.of_enum (BatList.enum vars)
+
+let run_symbolic config ~argument0 ?(arguments=[]) ?(vars=[]) colis =
   let open Symbolic in
   let clause_to_state root clause =
     let cwd = Constraints.Path.Abs [] in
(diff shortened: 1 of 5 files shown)


src/colis.ml
```diff
@@ -98,11 +98,18 @@ let colis_to_file filename colis =
 
 (* Interpret *)
 
-let run ~argument0 ?(arguments=[]) colis =
+let mk_concrete_var_env =
+  let open Concrete.Env in
+  List.fold_left
+    (fun env (var, val_) -> mixfix_lblsmnrb env var val_)
```
Niols: Uh. What is this function name?

src/colis.ml
```diff
@@ -156,7 +163,10 @@ type symbolic_config = {
   stack_size: int;
 }
 
-let run_symbolic config ~argument0 ?(arguments=[]) colis =
+let mk_symbolic_var_env vars =
+  BatMap.of_enum (BatList.enum vars)
```
Niols: Sadly, using Seq would drop the support for OCaml < 4.07 :-(

src/colis_cmd.ml
```diff
@@ -66,21 +80,24 @@ let speclist =
...
+    "--realworld",                 Set real_world,                " Use system utilities in concrete execution";
     "--symbolic-fs",               String set_symbolic_fs,        "FS Name of the initial symbolic filesystem in symbolic execution (values: empty, simple, fhs, default: empty)";
     "--prune-init-state",          Set prune_init_state,          " Prune the initial state in symbolic execution";
     "--loop-limit",                Int ((:=) loop_limit),         sprintf "LIMIT Set limit for symbolic execution of while loops to LIMIT (default: %d)" !loop_limit;
     "--stack-size",                Int ((:=) stack_size),         sprintf "SIZE Set the stack size for symbolic execution to SIZE (default: %d)" !stack_size;
     "--print-states",              String ((:=)print_states_dir), "DIR Save symbolic states as dot files in directory DIR";
     "--fail-on-unknown-utilities", Set fail_on_unknown_utilities, " Unknown utilities kill the interpreter";
+    "--",                          Rest add_argument,             "ARG... Pass all further arguments directory to the CoLiS interpreter";
```
Niols: "directory"? "directly", maybe?

src/colis.ml
benozol: Isn’t it beautiful? That’s what you get when you define a ternary operator `([←])` in Why3 and export to OCaml (left-bracket, less, minus, right-bracket).

src/colis_cmd.ml
benozol: Indeed!

src/colis.ml
Niols: Uh.
.
.
.

Niols: Can't we do anything about that? In the driver maybe? Also, what about "mixfix"?

benozol: I think “mixfix” means ternary. Not sure if we can just change the name. (Defining it in the driver would erase it.)

Niols: Then maybe we could have somewhere `let update_env = mixfix_slkfghklkesjbfgzkjh` and a comment? That would make the code easier to read in OCaml, I believe.

Niols: Maybe just at the beginning of the file.

benozol: Good idea.

Niols: :-)

benozol:
(BTW yesterday I tried to speed up the CI by only compiliing the project for all supported ocaml versions, and install the provers and replay the proofs only for the latest supported version. It didn’t work out because I have too little understanding of travis, but in case you have an idea: it’s in branch [`faster-ci`](https://github.com/colis-anr/colis-language/blob/faster-ci/.travis.yml))

PR #81, opened 2019-02-11
Refactor semantics and fix variable environment

benozol:
Change of scope of this PR (was: under-approximation of symbolic execution, which is not realistic before my holidays)

- fix the `Not_found` exception we observed today in the stats
- large refactoring of the semantics, specifically a better separation between the semantics and the filesystem, which will allow proving the symbolic engine towards the semantics, that is shared with the concrete interpreter
- commented postconditions stating the over and under approximations of the symbolic engine


src/concrete/driver.drv
```diff
@@ -1,3 +1,16 @@
-module semantics.Semantics
-  syntax function interp_utility "Utilities.interp_utility %1 %2 %3"
+prelude "module IdEnv = Map.Make (String)"
+
+module semantics.Env
+  syntax type env "%1 IdEnv.t * %1"
```
benozol: The environment is now a pair of an OCaml map and a default value

src/concrete/semantics.mlw
```diff
@@ -234,43 +213,39 @@ module Input
...
-  }
-end
+  val function ([]) (env:env 'a) (id:identifier) : 'a
+    ensures { result = M.(env.model[id]) }
 
-module Filesystem
-  type filesystem = unit
-  let empty = ()
-end
+  let function get (env:env 'a) (id:identifier) : 'a =
```
benozol: This function is an alias for a mixfix operators that are defined in the driver, and can be used in the OCaml code

src/symbolic/symbolicInterpreter.mlw
```diff
@@ -34,104 +29,30 @@ module Filesystem
...
-  clone export semantics.AbstractState
-    with type filesystem = filesystem
-
-  use semantics.State as S
-  function sem_state (sta:state) : S.state = {
-    S.stdin = sta.stdin;
-    S.stdout = sta.stdout;
-    S.filesystem = ();
-  }
+  clone export semantics.Semantics
```
benozol: Here we instantiate the semantics of the symbolic engine with the symbolic filesystem

src/concrete/interpreter.mlw
```diff
@@ -9,32 +9,24 @@ module Filesystem
...
-  use list.Append
-  use list.Length
-  use list.Nth
-  use ref.Ref
-  use option.Option
-  use bool.Bool
-  use map.MapExt
-  use int.Int
-  use ref.Refint
+  clone export semantics.Semantics
```
benozol: Here we instantiate the semantics for the concrete interpreter

benozol:
Hey @Niols! 

All checks fail because of the yaml>=1.0 dependency – do we have to wait for an update of the opam docker image or can we do anything about it?

Niols:
Hey, they fail on master too because of Morbig >= 0.10. It's annoying to have to wait for that long. Do you have an idea of an alternative ?

I could set up OPAM in the Docker image so that it uses distant package repositories rather than the built-in one? But that's bad for Travis' load and will probably result in a longer CI.

benozol:
No hurry! I was just wondering about the reason or a simple workaround (and signal my return :) )

Niols:
Welcome back :-)
It's a bit frustrating to have to wait for one whole month each time with the CI failing…

benozol:
OK if the update of the OPAM image takes *a month*, we should indeed think about workarounds. (Alas, I don’t know enough docker to understand your proposal but please go ahead. We could gain some time by replaying the proofs only in one ocaml version, as I tried in https://github.com/colis-anr/colis-language/commit/2bca73f7e21c4a27459ac2652252a73832c9e17f.)

benozol:
@Niols does commit c750f78 make sense to you?

Niols:
It makes sense. I didn't know that we directly had access to the git repository. But if it is indeed the case, then this commit should fix it.

benozol:
OK I think I can take responsabilit for this

PR #86, opened 2019-02-14
Implement CPU time limit for symbolic execution/constraint solver

claudemarche:


lib/constraints/src/implementation/naive/engine.ml
```diff
@@ -58,6 +58,7 @@ let normalize ?(limit=50) (disj : Disj.t) : Disj.t =
 
 let simplify (disj : Disj.t) : Disj.t =
   Log.debug (fun m -> m "Simplifying");
+  Log.check_cpu_time_limit ();
```
benozol: That seems like a good location to check, indeed! BTW I am working on changing codomain of the variable environment to pairs of strings and export flags – I hope to send a PR before the evening

Niols: Yes, but why in `Log`?

claudemarche: It was not easy for me to understand the modules structure in lib/constraints. Feel free to move it somewhere else.

Niols:
Fair enough!

PR #87, opened 2019-02-14
Exported variables

benozol:
- [x] Add export to syntax
- [x] Specify export semantics
- [x] Add export in conversion from shell to colis
- [x] Add test for concrete interpreter
- [X] Add flag in variable environment to reflect the status exported/not exported
- [X] Filter variables when calling a utility
- [x] Update concret interpreter
- [x] Update symbolic interpreter
- [x] Update proofs for concrete interpreter


benozol:
@Niols I think this is getting into shape. Two questions
1. Could you implement the import from Shell? (and push to this PR if you like)
2. Any fine idea how to test the export of variables?

benozol:
@claudemarche proposed to test the export of variables by implementing a dummy utility acting on a specific variable (e.g. printing it). Running this test in a normal shell, however, would require the same utility to be provided as an external shell script on the PATH.

Testing exported variables in CoLiS remains awkward, and this points to the problem: there is no interaction with the rest of the language, because the behaviour of unexported variables differs only for utilities, which are not specified in CoLiS. In addition, I suspect that utilities act only upon a limited set of variables (PATH, LANG, etc.), which are exported by default. Which would make our previous semantics of exporting everything valid. @Niols, any insights from POSIX? Let’s discuss again on Wednesday before merging?

Niols:
Hum,

Maybe we could "simply" implement the `env` utility? Actually, simply its simple version without arguments. This is basically Claude's idea except it's in the standard and will be present on the systems for comparison. Since the order in which thime environment is printed isn't specified, we would also probably need to implement `sort` without arguments.

I'm sorry I didn't take time to look into this PR. We're trying to finalize a journal version of an article before next Sunday. But yes, let's discuss about that wednesday.

benozol:
env+sort sounds good – hear you wednesday.

benozol:
Alright – now the tests obviously fail because the import of the builtin `export` is not yet supported.

Niols:
env+grep seems also pretty nice.
I finally took the time to convert `export A` and `export A=...` from Shell!

benozol:
Alright, then this is ready I guess. Any final words?

Niols:
There's an old Earth saying, Captain. A phrase of great power and wisdom and consolation to the soul in times of need.

Niols:
Allons-y!

benozol:
Thank you this seems quite appropriate.

PR #92, opened 2019-04-03
Introduce a combinator to combine utilities in sequence.

yurug:
- Offer a combinator to define a sequence of utilities.
- Use it to implement `rm` with multiple arguments.

### DIFF
--- a/src/symbolic/utilitiesSpecification.ml
+++ b/src/symbolic/utilitiesSpecification.ml
@@ -5,8 +5,6 @@ open Semantics__Buffers
 
 type utility = state -> (state * bool) list
 
-let identity = fun s -> [(s, true)]
-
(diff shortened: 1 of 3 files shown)


src/symbolic/utilitiesSpecification.ml
```diff
@@ -5,6 +5,14 @@ open Semantics__Buffers
 
 type utility = state -> (state * bool) list
 
+let identity = fun s -> [(s, true)]
```
benozol: Parametrize with boolean value (like `SymbolicUtilities.return: bool → utility`) and use also in `interp_true`, `interp_false`?

benozol:
There is now `compose_strict` and `compose_non_strict`

PR #93, opened 2019-04-04
Symbolic interpretation of programs accepts set of initial state

benozol:


src/colis.ml
```diff
@@ -161,70 +161,75 @@ let print_symbolic_state fmt ?id sta =
...
-    let root = Constraints.Var.fresh ~hint:"r" () in
-    let clause = FilesystemSpec.compile ~root config.fs_spec in
-    Clause.(add_to_sat_conj clause true_sat_conj) |>
-    List.map (clause_to_state root) |>
-    List.map run_in_state
+  let normals, errors, failures =
+    let loop_limit = Z.of_int config.loop_limit in
+    let stack_size = Z.of_int config.stack_size in
+    let inp = { Input.empty with argument0 } in
+    Interpreter.interp_program loop_limit stack_size inp (BatSet.of_list stas') colis
```
benozol: @Niols I think this is actually the function you want to use to continue symbolic execution (for example first `pre` then `post`). Would this work? It now accepts a set of initial symbolic states.

Niols: I indeed need more granularity than just `run_symbolic`. Could we maybe split all this wrapper around `interp_program` in several small functions and expose them? So that `run_symbolic` would basically become a function of just a few lines that an API user can copy and tweak easily?

src/colis.ml
```diff
@@ -160,71 +188,44 @@ let print_symbolic_state fmt ?id sta =
...
-  let print_symbolic_state label ctr fmt sta =
-    let id = sprintf "%s-%d" label !ctr in
-    incr ctr;
-    fprintf fmt "- @[%a@]@\n" (print_symbolic_state ~id) sta
+  let root = Constraints.Var.fresh ~hint:"r" () in
+  let disj = compile_fs_spec ~root fs_spec in
+  let stas = List.map (to_state ~prune_init_state:config.prune_init_state ~root) disj in
+  let stas' = List.map (to_symbolic_state ~vars ~arguments) stas in
+  let normals, errors, failures =
+    interp_program ~loop_limit:config.loop_limit ~stack_size:config.stack_size ~argument0
```
benozol: @Niols Short enough? Let me know what’s missing or unhandy for running the scripts.

benozol: This has not bee outdated but the following lines

https://github.com/colis-anr/colis-language/blob/8909f0551fec405b2ccd40e48f2c116d6c328410/src/colis.ml#L232-L237

Niols: This looks really great.

benozol: OK then I will wait for the CI and merge. BTW it also adds arguments `--add-symbolic-fs FILE` to the CLI, with an example file `simple.fs`.

PR #101, opened 2019-05-02
Prove over-approximation of symbolic interpreter

benozol:


benozol:
The proofs for over-approximation of the symbolic engine are complete, but there are two remaining  issues with Why3:

- Extraction fails with an anomaly. The symbolic interpreter is therefore necessarily a case of *I have only proved it correct, not tried it!*

- Implementation problem in the concrete interpreter https://github.com/colis-anr/colis-language/blob/ce81f069b3f04e405211ff32146d8ad501093087/src/concrete/interpreter.mlw#L343-L348

PR #102, opened 2019-05-16
Spring cleanup

benozol:
Here comes a little cleanup. Among others it removes the use of batteries from the symbolic engine, with two implications:
1. There are still some uses of batteries in the constraint solver. @Niols can we get rid of them?
2. The symbolic engine represents state sets as lists. We should instead use a datastructure with a more efficient append operation. Any ideas? Just an appendlist?


Niols:
1. Just took care of it.
2. What about lists of lists, as we mentioned?

benozol:
> 2. What about lists of lists, as we mentioned?

Well, concat of two list of lists is still a concat.

However, I did some experiments comparing simple implementations of data structures with a cheap concat operations, and reasonable map, filter, and partition. I found the fasted construct to be `type concat_list 'a = List (list 'a) | Concat (t 'a) (t 'a)`, which just delays all concatenations it is either mapped/filtered/etc or converted to a list. I am currently working on the proofs of its correctness and will push to this PR when it is ready

Niols:
Oh, if that's indeed a concatenation that you want and not adding the elements of a list, they you, list of list is a bad choice. Sequences trivially have the constant time concatenation. But then you pay the price of the continuations.

Niols:
Concat lists seem quite independent from the symbolic engine, don't they? Maybe they could go in an external library, in `lib/`, for instance? That would make the extraction a bit more tedious however.

I'm thinking of this because they might prove useful in the efficient constraint solver as well. Am I right to think that they make the `map_flatten` function more efficient than usual linked lists?

This is in no way urgent and can be done afterwards, but maybe we should discuss this a little bit on tuesday?

benozol:
That might be a good idea. And `map_flatten` is already available as `collection.ConcatList.bind` but more efficient than `concat (map f l)`. But I don’t think it is more efficient than `fold_left rev_append (rev_map f l) []`.

Niols:
Well, it would be interesting to benchmark this: concat lists against regular linked lists. You might have a speedup by avoiding constructing intermediary lists.
— Niols

PR #105, opened 2019-06-07
Separate traces and errors

### LINKED ISSUE #103: Utilities traces should not appear in stdout
The debugging traces of utilities are printed in the stdout of the interpreter. This changes the semantics of the script and this is therefore unsound.

We could extend the state of the interpreter with a log to put such debugging information, right?

benozol:
With this PR, the state of the interpreter(s) contains two output buffers: just the stdout, and another one called `log` that combines  stdout, trace messages of symbolic utilities (prefixed with `[DBG]`) and error messages (prefixed with `[ERR]`).

Closes #103


src/symbolic/symbolicUtility.ml
```diff
@@ -106,34 +109,34 @@ let failure ?error_message () =
...
-  |> List.map (quantify_over_intermediate_root state)
+  Clause.add_to_sat_conj case.spec sta.filesystem.clause
+  |> List.map (quantify_over_intermediate_root sta.filesystem)
   |> List.flatten
-  |> List.map (apply_clause_to_state state case root)
-
+  (* And now update the filesystem with the new clause and root *)
+  |> List.map @@ fun clause ->
+  let filesystem = {sta.filesystem with clause; root=root'} in
+  {sta with filesystem}, case.result
```
benozol: @Niols could you have a look if this function still looks reasonable? I tried to simplify.

Niols: I'm not sure if I understand every line of code, but it looks good!

benozol:
Update: keep a single output channel, `stdout`, and print the tracing and error output only when `--only-stdout` is not given.

PR #107, opened 2019-06-18
Add ands and ors to concrete syntax

### LINKED ISSUE #104: Extend colis with and/or operators on booleans
To improve the readability of generated colis programs, we should avoid the nesting of if-then-else which results from the encoding of or/and operators.

Could we extend colis with these boolean operators?

benozol:
closes #104

### DIFF
--- a/src/language/colisLexer.mll
+++ b/src/language/colisLexer.mll
@@ -4,35 +4,37 @@ open ColisParser        (* The type token is defined in colis_parser.mli *)
 exception LexerError of string
 
 let reserved_words =
-  [ "arg",      ARG ;
-    "begin",    BEGIN ;
-    "call",     CALL ;
-    "do",       DO ;
-    "done",     DONE ;
-    "else",     ELSE ;
-    "embed",    EMBED ;
-    "end",      END ;
-    "epip",     EPIP ;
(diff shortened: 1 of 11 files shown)


claudemarche:
That's good and already fine for the use of pretty-printing the colis codes generated by morsmall, but theer should also be parsing rules for those `and` and `or`

Niols:
Agreed. It would be nice to have that the parser is always capable of parsing what the pretty-printer outputs.

benozol:
Well - the meaning of the printer output is currently undefined: Any opinions about precedents of AND / OR? What about no precedence and just left-associativity?

claudemarche:
Good point: I suggest to put parentheses in the output, e.g. `[@(%a and@ %a)@]`

For parsing: AND have priority over OR, of course, and I suggest RIGHT associativity for both, not left

Niols:
In Shell, && and || have the same precedence. But it does not necessarilly make sense to have that in Colis.

— Nicolas Jeannerod

claudemarche:
Fine with me now

claudemarche:
Isn't there any choice "rebase and merge when ci succeeds?" too bad...

PR #111, opened 2019-06-26
Wrapper around Cmdliner to have easy command-line parsing

Niols:


Niols:
Considering how simple the commands are, I am not sure that spawning cmdliner makes sense. However, this gives us a nice POSIX- and GNU-compatible parser for free. Also, cmdliner is a very light dependency that is basically everywhere in the OCaml ecosystem. What do you think, @benozol @treinen?

If we like this way of doing things, I could go forward and rewrite all the argument parsing pieces of code for other utilities. (And maybe, while I'm at it, split the SymbolicUtilities module?)

benozol:
That would be a nice improvement over the current handling of arguments (+ I don’t fear dependencies : )

treinen:
I did not think of cmdliner but that might indeed be a good solution.

src/symbolic/symbolicUtilities.ml
```diff
@@ -260,15 +74,15 @@ let register () =
     "echo", interp_echo;
     "test", Test_utility.interpret ~bracket:false;
     "[", Test_utility.interpret ~bracket:true;
-    "touch", interp_touch;
-    "which", interp_which_full;
-    "silent-which", interp_silent_which;
     "update-alternatives", interp_update_alternatives;
     "dpkg", interp_dpkg;
   ];
   List.iter register [
```
benozol: When no utilities are defined in `SymbolicUtilities` anymore, this registration of all utilities can go into module `Colis`

Niols:
I already moved some of these. But do we really want a module for True and False, for instance? Or maybe I could put them all in a Basics module?

— Nicolas Jeannerod

benozol:
That's a good idea since there is already a precedent for a utility as a submodule (which.quiet)

Am 28. Juni 2019 09:14:43 MESZ schrieb Nicolas Jeannerod <notifications@github.com>:

Niols:
Well, this PR introduces the precedent so we could decide that we do not want it.

— Nicolas Jeannerod

Niols:
@benozol The last commits implement exactly this.

PR #115, opened 2020-01-17
Fix clause in SymbolicUtility.failure

benozol:
The `clause` in function `SymbolicUtility.failure` was just `true` and this PR corrects it to be `r = r'`.


Niols:
Oh indeed, how didn't we see that before? How didn't it show in weird behaviours?

benozol:
> How didn't it show in weird behaviours?

Because it was only triggered by weird usages of utilities, e.g., `mkdir ''`

Niols:
OK, makes sense!

PR #117, opened 2020-02-08
Functorize symbolic engine

### LINKED ISSUE #116: Functorize Symbolic Engine
As a first step towards plugging a transducer backend into the symbolic engine, we need to factorize the latter. I think the only thing the engine need are an abstract type for a filesystem (or a transformation of filesystem, the engine doesn't have to know the difference) and a dispatch function.

---

The Why3 code seems to be quite a lot constraints-aware:

https://github.com/colis-anr/colis-language/blob/a4da8ec5da50b29d934f9665e5374d187f0089e4/src/symbolic/driver.drv#L1-L6

But I think this is only used to define the type for a filesystem. And that type should probably be abstract for the symbolic engine, and defined in the constraint backend only?

https://github.com/colis-anr/colis-language/blob/a4da8ec5da50b29d934f9665e5374d187f0089e4/src/symbolic/symbolicInterpreter.mlw#L24-L28

---

The pathes from constraints are also used in Why3. Do we really need that? We could either provide a generic "path" notion, not included in the constraints. We could even only use strings, maybe? I don't know. It should go out of the constraints anyway.

https://github.com/colis-anr/colis-language/blob/a4da8ec5da50b29d934f9665e5374d187f0089e4/src/symbolic/driver.drv#L11-L12

---

I think the only other function needed would be the dispatch function, under a form or an other.

https://github.com/colis-anr/colis-language/blob/a4da8ec5da50b29d934f9665e5374d187f0089e4/src/symbolic/driver.drv#L18-L20

https://github.com/colis-anr/colis-language/blob/a4da8ec5da50b29d934f9665e5374d187f0089e4/src/symbolic/symbolicInterpreter.mlw#L324

benozol:
Fixes #116


benozol:
Hi @Niols , i have started working on functorizing the symbolic engine. In one go I also

- updated the Why3 version,
- replaced the old `mystring.String` module with the module now provided by Why3, and
- split specification of concrete interpretation of utilities from their abstract implementation

This is WIP because I haven't yet attempted to resurrect the proofs, but it should be a starting point to continue the functorization of the OCaml code base.

benozol:
We are getting closer (it compiles but I haven't updated the proofs). These are the functor arguments for parameterizing the symbolic engine and the utllity specifications:

https://github.com/colis-anr/colis-language/blob/5d1110291e3661010f85666ca9d881ec62060608/src/symbolic/symbolicUtility.ml#L5-L16

 @Niols what do you think?

Niols:
It looks good. Two questions/remarks though:

1. Why are there two types `FILESYSTEM.filesystem` and `CASESPEC.filesystem`?

2. Wouldn't that be the time to say that `apply_spec` only returns one filesystem, and that if your backend needs to introduce disjunctions, it should put them in the `filesystem` type? Basically, move the DNF situation from the engine to the backend. I am not convinced we should have made the engine aware of the fact that the constraints were always in DNF.

benozol:
> 1. Why are there two types FILESYSTEM.filesystem and CASESPEC.filesystem?

`FILESYSTEM` and `CASESPEC` are arguments to different (nested) functors. Stating `CASESPEC.filesystem` is required to define the signature of `apply_spec`. `FILESYSTEM` and `CASESPEC` can be implemented by a single module of course, e.g. `UtilityConstraints.{Constraints,Transducers,Mixed}`

> 2. Wouldn't that be the time to say that apply_spec only returns one filesystem

Sure that would be possible and would enable the sharing of concrete parts of the state between the results of `apply_spec`. But we would have disjunctions on two levels (e.g., to implement a utility that succeeds or fails depending on the filesystem) :/ so is this "only" an optimisation issue?

Niols:
2. Well, we already have disjunctions on two levels: the logic level and the engine+specification level. It would make sense to recover that in the code, I imagine. Also, on the constraints level, I am now designing a new system that doesn't introduce disjunctions ever. So I will always give you a singleton list. I guess transducers will also probably always give a singleton list.

Niols:
2. Note however that it should only change from list to option, since a case spec might be impossible to apply. So this is really not so much a matter of optimisation.

benozol:
> I am now designing a new system that doesn't introduce disjunctions ever. So I will always give you a singleton list. I guess transducers will also probably always give a singleton list

In these cases there's of course no sense for `apply_spec` to return a list (indeed instead of an option). Should we postpone this until we #117-#125 are ready and/or the new constraint system is in place?

benozol:

PR #119, opened 2020-02-17
Root variables in case specification

benozol:
This PR basically changes the type of symbolic case specifications from `Clause.t` to `Var.t → Var.t → Clause.t` (= `type case_spec`).

This is required to isolate the dealing with constraint variables in `under_specifications` in one function (`SymbolicUtility.add_spec`), as required for #116.

It allows a natural case specification `noop` (= `Clause.eq`) which avoids bugs like #115


benozol:
@Niols is there any code that depends on the fact that all states resulting from the interpretation of a utility in state share the same new root variables, or was this just premature optimization :) ?

Niols:
No, nothing depends on that. The only important thing is that we get access to the new root of one state, either through a closure or by storing that root in the state.

PR #122, opened 2020-02-21
Mixed case specs

benozol:
This PR allows for mixing constraint-based and transducers-based case specifications for utilities (transducers only as dummy).

This minimizes the effort to implement transducers-based specifications utilities because  all the concrete evaluation parts of utilities, including parsing of utility parameters etc., are shared.

Mixed specifications are currenlty only used in the implementation of utility `mkdir` (line 25 ff):

https://github.com/colis-anr/colis-language/blob/72e28559ba37bc1e67b7825e197ba563b30c716c/src/symbolic/utilities/mkdir.ml#L21-L30

The adaption of the other utilities (from `SymbolicUtility.ConstraintsCompatibility` to `SymbolicUtility.Mixed`) is straightforward: not providing the parameter `~transducers` to `case_spec` induces incomplete behaviour.


Niols:
So I thought we would have different sets of specifications. But this actually sounds nice. We might discover this a bit rigid in the future, but I am not sure of it and this is still a nice step forward (let's not go to premature optimisation, the changes are big enough as they are).

I validate the idea and will now review the code.

> The adaption of the other utilities (from `SymbolicUtility.ConstraintsCompatibility` to `SymbolicUtility.Mixed`) is straightforward: not providing the parameter `~transducers` to `case_spec` induces incomplete behaviour.

Then I don't really understand why you would use `()` in `mkdir`, for instance? Why not just not provide `~transducers`? Also, I imagine that the same holds for `~constraints` (in case there is something transducers specify and constraint don't; I should also have my answer to this question when reviewing)?

src/symbolic/symbolicInterpreter.mlw
```diff
@@ -34,7 +35,7 @@ module Interpreter
       scope import Arg
         (* TODO Remove `function` from this and the below reformulation, use predicate in specifications below *)
         val function sym_interp_utility (params: (utility_context, identifier, state)) : Cn.t (state, result bool)
-          ensures { forall res. interp_utility params res <-> Cn.mem res result }
+          ensures { forall res. interp_utility params res -> Cn.mem res result }
```
Niols: I am not sure I understand why this was an equivalence, and why this isn't anymore?

src/symbolic/symbolicUtility.ml
```diff
@@ -319,34 +344,146 @@ module SymbolicImplementation = struct
...
+  type case_spec = {
+    constraints: unit -> ConstraintsImplementation.case_spec;
+    transducers: unit -> TransducersImplementation.case_spec;
   }
 
-  type filesystem = symbolic_filesystem = {
-    root: Var.t;
-    clause: Clause.sat_conj;
-    root0: Var.t option;
+  let case_spec ?transducers ?constraints () : case_spec =
```
Niols: This answers my question about constraints being optional. Good choice.

benozol:
> Then I don't really understand why you would use () in mkdir, for instance?

`()` is only a dummy example of a dummy case specification for transducers. Not providing the `~transducers` to `case_spec` at all will induce incomplete behaviour for transducers even in cases that are `success` or `error`.

src/symbolic/symbolicInterpreter.mlw
benozol: That was erroneous before – we are relying only on over-approximation of sym_interp_utility to prove the over-approximation of the engine

benozol:

PR #124, opened 2020-02-22
Clean interface of module Colis

benozol:


Niols:
Not much to say here, I think the result indeed looks clean. Thank you for the update in colis-batch, by the way.

benozol:

PR #126, opened 2020-02-25
Shadow aware printing

Niols:
This PR:

- Cleans up a bit the internal interface of constraints.

- Makes the constraint printer more aware of how "shadows" work. They are a mechanism to hide nodes in the printed output. Without it, with all the nodes introduced by the FHS and the unpacking, the printed output would be unreadable. The previous printer was hiding too many nodes, giving weirdly looking (and sometimes unexploitable) results. This one gives much better results.

@benozol I don't need your input on that. However, I wanted to check first that you had not changed anything wrt the constraints? I'm a bit lost in all the changes and I wouldn't want to break your code for such a minor change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ services:
 ## are in fact tags for the ocaml/opam2 Docker image that we pull.
 
 env:
+  - TAG=4.09
+  - TAG=4.08
   - TAG=4.07
   - TAG=4.06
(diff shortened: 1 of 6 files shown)


Niols:
The CI seems quite unhappy with Why3's replaying of proofs. I guess it comes from the commit I branched from on master, and all should go well later with the merges?

PR #132, opened 2020-03-26
Rework Dockerfile and improve CI

### LINKED ISSUE #127: Drop support for OCaml < 4.08.0?
It is not necessarily my habit to drop support for "old" versions of OCaml as soon as possible. However, I think it has been several times in this project that I tried to use 4.08 features. In the meantime, OCaml 4.08.0 has reached Debian stable. And unstable/testing are in 4.08.1. Could we and should we drop older versions of OCaml?

@benozol @treinen @claudemarche Any opinion?

### LINKED ISSUE #128: Avoid replaying the proofs in all contexts for a faster CI?
Everything is in the title. Currently, during the CI, we replay the proofs in all contexts (that is in all versions of OCaml). This is probably pointless, isn't it? Couldn't we just replay the proofs in one context?

@benozol What do you say? I can make the change if that sounds good to you.

Niols:
This PR rewrites the whole Dockerfile. It uses so called "multi-stage builds" that allow to run only part of a Dockerfile, and to define some form of graph dependencies (instead of the ordered dependencies usual Dockerfile use).

This allows (both are easily possible separately but not together):
- to only build and test the artifact without installing the provers or replaying the proofs
- to cache the installation of provers even if there are changes in files.

After this change was done, I used it to:

- configure travis to only run tests by default, and replay proofs only on one job (fix #128)
- add CI on OCaml 4.10; remove CI on OCaml 4.05, 4.06, 4.07 (fix #127)
- cleanup makefile
- cleanup travis.yml

As a consequence the two first points, the CI is now much faster. It consists in only 3 jobs of around 10, 10 and 20 minutes. It was before made of 5 jobs of 20 minutes each. When submitting a PR, for instance, Travis runs twice all the jobs with a limit of 5 jobs per organisation, which means that it now takes only 20 minutes (with actually 2x10 minutes saved that can be used on other jobs) against 40 before.

@benozol Does that sound good to you? (If yes, I guess you can just merge this directly.)

---

The following table gives an idea of the time spent (on my laptop) in each stage of the Dockerfile.

| Stage                       | Depends on                                          | Time     |
|----------------------------|-----------------------------------------------------|-----------|
| `basis`                      | the `TAG` argument                               | 1-3 min |
| `basis-with-deps`     | `basis`, the file `colis-language.opam`  | 8 min    |
| `builder`                   | `basis-with-deps`, all the other files       | 1 min    |
| `tester`                     | `builder`                                                 | 1 min    |
| `basis-with-provers` | `basis`                                                   | 3 min    |
| `prover`                    |`basis-with-deps`, `basis-with-provers` | 8 min    | 

The following table gives an idea of the time spent (on my laptop) for full builds and rebuilds of the targets `builder` and `prover` after having changed the `colis-language.opam` file or any otherfile.

<table>
    <thead>
        <tr>
            <th>Target</th>
            <th>Rebuild after change to</th>
            <th>Uses cache for</th>
            <th>Time</th>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td rowspan="3"><code>tester</code></td>
            <td></td>
            <td></td>
            <td>12 min</td>
        </tr>
        <tr>
            <td><code>colis-language.opam</code></td>
            <td><code>basis</code></td>
            <td>9 min</td>
        </tr>
        <tr>
            <td>any other file</td>
            <td><code>basis</code>, <code>basis-with-deps</code></td>
            <td>2 min</td>
        </tr>
        <tr>
            <td rowspan="3"><code>prover</code></td>
            <td></td>
            <td></td>
            <td>22 min</td>
        </tr>
        <tr>
            <td><code>colis-language.opam</code></td>
            <td><code>basis</code>, <code>basis-with-provers</code></td>
            <td>16 min</td>
        </tr>
        <tr>
            <td>any other file</td>
            <td><code>basis</code>, <code>basis-with-deps</code>,<br/><code>basis-with-provers</code></td>
            <td>9 min</td>
        </tr>
    </tbody>
</table>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,18 @@
-language: c
-sudo: required
 services:
   - docker
 
-## We build on as many OCaml version as we can support. These versions
-## are in fact tags for the ocaml/opam2 Docker image that we pull.
-
 env:
-  - TAG=4.09
-  - TAG=4.08
-  - TAG=4.07
-  - TAG=4.06
-  - TAG=4.05
+  global:
+    - TARGET=tester
+  jobs:
+    - TAG=4.10 TARGET=prover
+    - TAG=4.09
+    - TAG=4.08
 
 script:
   - docker build
       --build-arg "TAG=$TAG"
       --build-arg "SWITCH=$SWITCH"
       --tag colisanr/colis-language:$TRAVIS_BRANCH
+      --target "$TARGET"
       .
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,82 +1,85 @@
-## =============================== [ Basis ] =============================== ##
+## ============================== [ Basis ] ================================= ##
 
 ARG TAG=latest
 ARG IMAGE=ocaml/opam2:$TAG
 
-FROM $IMAGE
+FROM $IMAGE as basis
 
-MAINTAINER Nicolas Jeannerod
+ENV LANGUAGEDIR=/home/opam/colis-language
+ENV PROVERSDIR=/home/opam/provers
 
-ENV OPAMFILE=colis-language.opam
+ENV Z3="z3-4.8.7"
+ENV CVC4="cvc4-1.6"
+ENV ALTERGO="alt-ergo.2.3.0"
+
 ARG SWITCH=
-
 RUN [ -z "$SWITCH" ] || opam switch create "$SWITCH"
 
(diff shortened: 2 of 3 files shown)


benozol:
That sounds very good indeed.

(And it's always the right time for better CI ☺ – how much time have I spent waiting for it!)

PR #133, opened 2020-04-06
Plug transducers backend into CLI of symbolic interpreter

benozol:
Add an option `--symbolic-backend` to the command line interface to select either the constraints-backend or the transducers-backend.

### DIFF
--- a/src/colis_cmd.ml
+++ b/src/colis_cmd.ml
@@ -3,6 +3,7 @@ open Colis
 
 type source = Colis | Shell
 type action = Run | RunSymbolic | PrintColis | PrintShell
+type backend = ConstraintsBackend | TransducersBackend
 
 let (||>) f g x = f x |> g
 
@@ -16,6 +17,23 @@ let get_action, set_action =
     match !action with
     | None -> action := Some new_action
     | Some _ -> raise (Arg.Bad "only one action among --run, --run-symbolic, --print-colis and --print-shell can be specified"))
+
+let get_symbolic_backend, set_symbolic_backend =
+  let symbolic_backend = ref None in
+  let parse_symbolic_backend = function
+    | "CONSTRAINTS" -> ConstraintsBackend
+    | "TRANSDUCERS" -> TransducersBackend
+    | _ -> raise (Arg.Bad "invalid backend (should be constraints or transducers)") in
+  (fun () ->
+     match !symbolic_backend with
+     | None -> ConstraintsBackend
+     | Some backend -> backend),
+  (fun new_backend ->
+     if get_action () <> RunSymbolic then
+       raise (Arg.Bad "symbolic backend can only be selected after specifying --run-symbolic");
+     match !symbolic_backend with
+     | None -> symbolic_backend := Some (parse_symbolic_backend new_backend)
+     | Some _ -> raise (Arg.Bad "symbolic backend can only specified once"))
 
 let get_source, set_source =
   let source = ref None in
@@ -90,7 +108,8 @@ let speclist =
   let open Internals.Options in
   align [
     "--run",               Unit (set_action Run),         " Concrete execution (default)";
-    "--run-symbolic",      Unit (set_action RunSymbolic), " Symbolic execution";
+    "--run-symbolic",      Unit (set_action RunSymbolic), " Symbolic execution with constraints backend";
+    "--symbolic-backend",  String set_symbolic_backend,   "BACKEND Set backend of symbolic execution (constraints or transducers)";
     "--shell",             Unit (set_source Shell),       " Use the shell parser (default)";
     "--colis",             Unit (set_source Colis),       " Use the colis parser" ;
     "--external-sources",  Set_string external_sources,   "DIR Import absolute sources from DIR";
@@ -112,10 +131,11 @@ let speclist =
 let usage =
   sprintf
     ("Usage: %s <action> <syntax-opts> [--var VAR=VAL ...] FILE [--] [ARG...]\n"^^
-     "       <action>: [--run <opts> <concrete-opts> | --run-symbolic <opts> <symbolic-opts> | --print-colis | --print-shell]\n"^^
+     "       <action>: [--run <opts> <concrete-opts> | --run-symbolic [<opts> <constraints-opts>|<transducers-opts>] | --print-colis | --print-shell]\n"^^
      "       <opts>: [--unknown-behaviour EXCEPTION|UNSUPPORTED|ERROR]\n"^^
      "       <concrete-opts>: [--realworld]\n"^^
-     "       <symbolic-opts>: [--symbolic-fs FS] [--prune-init-state] [--loop-boundary] [--print-states DIR]\n"^^
+     "       <constraints-opts>: [--symbolic-backend CONSTRAINTS] [--symbolic-fs FS] [--prune-init-state] [--loop-boundary] [--print-states DIR]\n"^^
+     "       <transducers-opts>: --symbolic-backend TRANSDUCERS [--symbolic-fs FS]\n"^^
      "       <syntax-opts>: [--shell [--external-sources DIR] | --colis]")
     Sys.argv.(0)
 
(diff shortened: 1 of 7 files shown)


benozol:
Hi @Niols what do you think? Would be nice to have this in place for tomorrow's visio.

src/colis_cmd.ml
```diff
@@ -17,6 +18,23 @@ let get_action, set_action =
     | None -> action := Some new_action
     | Some _ -> raise (Arg.Bad "only one action among --run, --run-symbolic, --print-colis and --print-shell can be specified"))
 
+let get_symbolic_backend, set_symbolic_backend =
+  let symbolic_backend = ref None in
+  let parse_symbolic_backend = function
+    | "CONSTRAINTS" -> ConstraintsBackend
+    | "TRANSDUCERS" -> TransducersBackend
+    | _ -> raise (Arg.Bad "invalid backend (should be constraints or transducers)") in
```
Niols, line 26: This doesn't seem case insensitive but it probably should be? Or at least the documentation should not mention the cases in lowercase.

src/colis_cmd.ml
```diff
@@ -3,6 +3,7 @@ open Colis
 
 type source = Colis | Shell
 type action = Run | RunSymbolic | PrintColis | PrintShell
+type backend = ConstraintsBackend | TransducersBackend
```
Niols, line 6: Shouldn't we move the whole command line parsing into a specific "Config" or "Options" module? (Maybe not for this PR.)

src/colis_cmd.ml
```diff
@@ -112,10 +131,11 @@ let speclist =
 let usage =
   sprintf
     ("Usage: %s <action> <syntax-opts> [--var VAR=VAL ...] FILE [--] [ARG...]\n"^^
-     "       <action>: [--run <opts> <concrete-opts> | --run-symbolic <opts> <symbolic-opts> | --print-colis | --print-shell]\n"^^
+     "       <action>: [--run <opts> <concrete-opts> | --run-symbolic [<opts> <constraints-opts>|<transducers-opts>] | --print-colis | --print-shell]\n"^^
      "       <opts>: [--unknown-behaviour EXCEPTION|UNSUPPORTED|ERROR]\n"^^
      "       <concrete-opts>: [--realworld]\n"^^
-     "       <symbolic-opts>: [--symbolic-fs FS] [--prune-init-state] [--loop-boundary] [--print-states DIR]\n"^^
+     "       <constraints-opts>: [--symbolic-backend CONSTRAINTS] [--symbolic-fs FS] [--prune-init-state] [--loop-boundary] [--print-states DIR]\n"^^
```
Niols, line 137: Why is the loop boundary specific to constraints? Also, if we are planning to run batch on transducers, it would be nice to have a way to print states there too. But we can also see that later.

src/colis_cmd.ml
```diff
@@ -144,15 +164,24 @@ let main () =
...
+      let fs_spec = get_symbolic_fs () in
+      match get_symbolic_backend () with
+      | ConstraintsBackend ->
+        let open SymbolicConstraints in
+        let config = {
+          prune_init_state = !prune_init_state;
+          loop_limit = !loop_limit;
+          stack_size = !stack_size;
+        } in
+        let res = run config fs_spec ~argument0 ~arguments ~vars program in
```
Niols, line 177: I'm a bit sad that the `run` commands don't look the same. I guess it would be annoying to split the config into "engine config" and "backend config"? What if the config of the engine was something like:

    type 'a config =
        { backend : 'a ;
          loop_limit : int ;
          stack_size : int }

?

src/symbolic/filesystemSpec.mli
```diff
@@ -26,7 +24,7 @@ val add_channel : in_channel -> t -> t
 
 (** Compile the specification of a filesystem into a feature constraint at the given root
     variable.*)
-val compile : Var.t -> t -> Clause.t
+val compile_constraints : Colis_constraints.Var.t -> t -> Colis_constraints.Clause.t
```
Niols, line 27: I guess we want the same thing for transducers, but for now we don't know how to compile a filesystem into a transducer?

src/symbolic/symbolicUtility.ml
```diff
@@ -92,7 +95,7 @@ module MakeInterpreter (Filesystem: FILESYSTEM) = struct
   let is_registered = Hashtbl.mem table
 
   let () =
-    (* TODO Register all known POSIX utilities:
+    (* TODO Register all essential commands as:
```
Niols, line 98: If I'd want to be picky, I'd say "essential utilities", just removing the POSIX (as Ralf reminds us regularly, we don't care about POSIX utilities, but just about the ones accessible in Debian). No strong feeling on this though.

src/symbolic/symbolicUtility.ml
```diff
@@ -92,7 +95,7 @@ module MakeInterpreter (Filesystem: FILESYSTEM) = struct
...
        [incomplete ~utility "not implemented"] *)
     (* TODO Register all known unknown POSIX utilities, if any:
```
Niols, line 100: I guess you'd have to remove the "POSIX" here too right?

src/symbolic/symbolicUtility.mli
```diff
@@ -235,6 +237,7 @@ module Constraints : sig
     with type filesystem = ConstraintsImplementation.filesystem
   include module type of MakeInterpreter (ConstraintsImplementation)
   include module type of MakeSpecifications (ConstraintsImplementation)
+  val filesystems : prune_init_state:bool -> FilesystemSpec.t -> filesystem list
```
Niols, line 240: I guess here the `filesystems` functions could have the same type but they would take the backend's configuration as first argument. Also, abstracting away the backend config would allow to change it later on without affecting the whole interface of the program, and the symbolic engine, etc.

Niols:
> Hi @Niols what do you think? Would be nice to have this in place for tomorrow's visio.

Hey. Sorry, I had a talk yesterday and I did not check my GitHub notifications until this morning.

src/colis_cmd.ml
benozol: OK, made it case-insensitive

src/colis_cmd.ml
benozol: Fixed for loop and stack limits. Printing transducer states should be implemented when we know what they are, I agree.

src/colis_cmd.ml
benozol: I thought that's just module `Colis_cmd`

src/colis_cmd.ml
benozol: Fully agreed – I separated `type sym_config = { loop_limit; stack_size; filesystem_spec }` and backend configs in `SymbolicUtility.Constraints` and `SymbolicUtility.Transducers`.

src/symbolic/filesystemSpec.mli
benozol: Exactly

src/symbolic/symbolicUtility.mli
benozol: Good point, done!

src/symbolic/symbolicUtility.ml
benozol: OK!

src/symbolic/filesystemSpec.mli
Niols: OK, we'll see that with them in time then.

Niols: Should we open issues for that kind of features-that-are-going-to-be-missing-eventually?

src/colis_cmd.ml
Niols: Oh yes, it's just if we wanted to split that to a separate module, but not necessarily. You're right though, `Colis_cmd` is basically just parsing command line and then calling `Colis`.

src/symbolic/filesystemSpec.mli
benozol: Yes, either by opening issues, or by adding the functions with an implementation `failwith "not implemented"` or so

PR #134, opened 2020-04-09
Add a concrete backend to the symbolic engine

benozol:


Niols:
So by "concrete backend", I guess you mean a "symbolic backend" that always returns only one state that is obtained by the concrete run of the command?

benozol:
Exactly. Added a comment for explanation.

